PR #2, opened 2017-04-25
Support for Chrome updated in the intro section

eva-seydl:
Our app supports from 8.1.53 Chrome OS.


eva-seydl:
Changing support for new Android devices

PR #11, opened 2017-04-28
Fixed spelling mistake...

Panzerbjrn:
Small mistake I caught...


lizap:
@Panzerbjrn - good catch! You'll see your fix live today. Thanks!

PR #15, opened 2017-05-01
Updated rds-support-config with SSO info and tweaked Connection Broker messaging

ChristianMontoya:
Added the SSO section for clarity on the scenario after receiving several requests from customers. Also, would you be able to check if the links resolve? I wasn't sure how to do that in this web experience.


lizap:
@ChristianMontoya - Christian, the links look ok, but I'll double-check them once I pull them into the private repo, where we get staging. Thanks!

PR #18, opened 2017-05-03
Escape angle brackets wrapping place holders

cparker4486:


lizap:
Thanks for fixing those brackets! You'll see the change go live this morning.

illfated:
(ref. **Appendix H: show hidden placeholders (ref fa23beb)** ) #5017

PR #29, opened 2017-05-09
Fixed incorrect formatting and removed string

chrisbrownie:
A paragraph was title formatted, removed that. A random GUID-looking string was present with some quote markers, removed that too.


lizap:
Hi, Chris - You've actually identified a merge conflict that slipped through our gates. I'm going to go ahead and merge your PR, but you'll see a change made before we publish the update - the paragraphs are actually dupes, so I'll get rid of one of them. Thanks for catching this!

PR #49, opened 2017-05-23
Adding POP and IMAP values for x-ms-client-application

psignoret:
Per https://support.microsoft.com/en-us/help/3107357/pop-imap-client-authenticate-fails-if-x-ms-client-application-in-the-ad-fs-claim-rule-is-set-to-microsoft.exchange.popimap


billmath:
Looks good.  Thanks!

#sign-off

PR #52, opened 2017-05-23
Modified guest OS support statement

scooley:
Hyper-V's guest support statement is the same as Microsoft's when server 16 shipped.  I believe Vista wasn't supported and Server 2008/Windows 7 had just stopped being supported but it may take some forensics.

https://support.microsoft.com/en-us/help/13853/windows-lifecycle-fact-sheet

https://support.microsoft.com/en-us/lifecycle/search for server OSes


lizap:
Thanks, Sarah! I've merged the PR and will get it pulled over to the private repo for publish later today.

PR #55, opened 2017-05-26
Fix and improve Windows commands

### LINKED ISSUE #50: Markdown issues with windows commands
Many of the windows commands have problems with markdown formatting. There are angle brackets that should be escaped, and some sections and notes that do not display properly because there isn't enough spacing. The capitalization also seems to have disappeared in many places. I'm guessing these problems resulted from when the documents were converted into markdown.

b-r-o-c-k:
I would like to help resolve #50 and improve the markdown for Windows commands.
This pull request contains the following changes:
- Added more whitespace to fix formatting problems
- Fixed capitalization
- Made all "Additional References" headers use h4 tags
- Added code blocks to all "Syntax" sections

If there are any other changes that need to be made please let me know.


lizap:
@b-r-o-c-k - Hi, Brock - I've asked one of the writers here to take a review pass through your changes. Assuming all the changes are right, which I assume they are, we'll get this into the pipeline ASAP.

Thanks for your help!

PR #65, opened 2017-06-02
Fixing path parameter on Export-PfxCertificate

linkhp:


Justinha:
This is already fixed and live at https://docs.microsoft.com/en-us/windows-server/virtualization/guarded-fabric-shielded-vm/guarded-fabric-prepare-for-hgs

Justinha:
This is fixed and live at https://docs.microsoft.com/en-us/windows-server/virtualization/guarded-fabric-shielded-vm/guarded-fabric-add-host-information-for-tpm-trusted-attestation

PR #85, opened 2017-06-12
Fixing typo in VirtualNetworkProperties

nwoodmsft:
In the example code there is a typo - NbVirtualNetworkProperties should be VirtualNetworkProperties


lizap:
Good catch! Thanks - you'll see your change live on Monday, 6/19.

PR #93, opened 2017-06-15
Line 87, wording

LucianFrango:
Wording didn't make sense. Updated to read better.


lizap:
Thanks for the fix - you'll see your change by 11 AM, 6/16.

PR #96, opened 2017-06-16
Fix usages of Vertical Line in certreq_1.md file

0xfeeddeadbeef:
Replace Vertical Line "|" characters when used as a pipe symbol inside tables with HTML entity `&#124;`
Parts of example commands after pipe were lost from rendered markdown.

Also replaced misspelled "aigid" with "algid" in example.


0xfeeddeadbeef:
I guess there are lot more places in this repo where Vertical Line is hiding text from rendered tables.
Just guessing. This might be a much larger problem.

PR #108, opened 2017-06-27
Update diskcomp.md

abcdlzy:
fix display issue


lizap:
Thanks for the fixes! You'll see them live on the site on 6/27/17.

PR #120, opened 2017-07-11
Storage Spaces UI shows NVDIMM-N bus type as unknown

kaushika-msft:
1. The "unknown" classification in UI causes confusion that NVDIMM-N storage may be a problem and hence using the content guidance, customers will be able to validate the bus type before calling Microsoft or OEMs.
2. This was reported by DellEMC during their testing of NVDIMM hardware with Windows Server 2016 storage spaces.


lizap:
Hi, Kaushik - I think you're missing a word here: It doesn't have any fuctionality loss or inability **to  in** creation of Pool, Storage VD.

Can you fix that? Then I'll merge and get this update published today.

PR #129, opened 2017-07-21
Update Deploy-Nano-Server.md

chnylen:
In the "Installing a Nano Server WIM" section, the page displays:

Apply the Nano Server image (adjust the path of the .wim file):  
  Dism.exe /apply-imagmediafile:.\NanoServer.wim /index:1 /applydir:n:\
   
which is wrong, and does not work. The correct dism command is:

Dism.exe /apply-image /imagefile:.\NanoServer.wim /index:1 /applydir:n:\


jaimeo:
Good catch of an unfortunate typo.  Please merge.

PR #130, opened 2017-07-22
Update Deploy-Nano-Server.md

chnylen:
The R in reboot for the wpeutil command must be capitalized in order for it to work.


lizap:
@chnylen - Hi, Chris - do you think we need to add a note that states that capitalization there matters?

Thanks, Liza

chnylen:
It would definitely be helpful, I don't see the harm in adding it.

PR #141, opened 2017-07-29
Updated SDN cert info

jamesmci:
Added info about adding another EDU OID to the SDN cert


jamesmci:
Ready to merge, thanks!

PR #149, opened 2017-08-10
Fix for contributors guide link and link to Hyper-V Integration services docs

rh072005:
Both links were returning 404s


lizap:
Thanks for these fixes! You'll see them live later on 8/10 or 8/11.

PR #200, opened 2017-09-24
Fix links and useless redirections.

KexyBiscuit:


lizap:
Thanks for catching all those links! We're taking a look now, particularly for any of the links where we need to include ( ) (they don't necessarily make it through the publishing build) and will merge your changes ASAP.

jaimeo:
Fixing the "outside" links to relative is great (thanks!), but "en-us" (or any other locale) should never be added (or even used in) the links that still have to go outside. This is so that users in other locales will be served localized versions automatically. Adding "en-us" forces on them the English version no matter what their settings are. There are exceptions (certain marketing links break this rule, though they shouldn't), but none of these are, as far as I can see.

KexyBiscuit:
@jaimeo Definitely you're right, I hate adding locale code into URL, but some Microsoft websites do. To keep it consistent with other pages in the library, I choose to add locale codes. This was a problem I had wanted to report, but after Microsoft released this new documentation site it's no longer a problem.

jaimeo:
There are other issues with various proposed link changes here. Besides 'en-us' (which we do not use unless the target link doesn't work without it--sometimes the case with marketing sites), it's not a good idea to replace FWlinks w/o careful research. FWlinks are used when the target is at risk of changing as further things are published; they are typically used in many source locations. In this way we can update the FWlink on the back end and have all of them work wherever they occur. "Fixing" one in this topic by changing it to a direct links breaks our ability to update the FWlink once in several places and results in the local link now being wrong.

I will have to research and test each of the link changes you propose, but I won't be able to get to that until certain nearby deadlines pass.

KexyBiscuit:
@jaimeo I'd like to know which kinds of link changes are safe, and which are not.

jaimeo:
I'm not sure I can enumerate for you all possible cases, as that requires inside knowledge about how we structure and manage content. Managing links is a hairy job, so we are very grateful when people can point out apparent link problems (like 404s), but it might be safest to just report those as Issues rather than trying to fix them directly. When you do report a link Issue, please always include which topic the problematic link appears in as well as the link itself and where you think it ought to be going vs. where it is going.

PR #207, opened 2017-10-02
Updating so Existing AD works with AAD DS

ChristianMontoya:
Adding guidance in the AAD DS + Existing doc so that the deployment will succeed. Currently, it defaults to a specific local IP address that may not be true in many or most deployments.

Still need to add files, but email me for those.


lizap:
Merging. Will add images in a new branch and then merge to private for publish.

PR #214, opened 2017-10-04
Clarifying the checkpoint support

kaushika-msft:
> [!NOTE]
> Only **Production Checkpoints** are supported on guests that run Active Directory Domain Services role (Domain Controller) or Active Directory Lightweight Directory Services role.


lizap:
Thanks, Kaushik - merging now. Due to a publishing freeze, you'll see this fix live at launch. Let me know if you need a staging link to verify the fix. - Liza

PR #215, opened 2017-10-04
Added DC feature to Availability and backup section

kaushika-msft:
Domain controller are supported on all the Windows Server guests but with production checkpoints only.


lizap:
Thanks, Kaushik - merging now. Due to a publishing freeze, you'll see this fix live at launch. Let me know if you need a staging link to verify the fix. - Liza

PR #218, opened 2017-10-05
add .NET/C# sample code and complete list of faults

cosmosdarwin:
They are retiring MSDN dev centers, so I need to merge this content in here. I reorganized the page so it fits in, and was pretty careful with the formatting. Also fixed some casing while I was at it. We can polish later, for now this just needs an appropriate-ish home.


lizap:
Thanks, Cosmos - I'll merge but see my comment on the other PR - it's going to be a bit before the changes hit live. - Liza

PR #219, opened 2017-10-05
add .NET/C# sample code and complete list of metrics

cosmosdarwin:
They are retiring MSDN dev centers, so I need to merge this content in here. I reorganized the page so it fits in, and was pretty careful with the formatting. Also fixed some casing while I was at it. We can polish later, for now this just needs an appropriate-ish home.


lizap:
Thanks, Cosmos - I'll merge this now. Because of a publishing freeze in advance of launch, though, you won't see the changes live for a bit. Let me know if you need a staging link to verify your changes. - Liza

PR #220, opened 2017-10-05
Honolulu: fixed deployment guide for Windows Server Core

uzzzerrrname:
Fixed `msiexec` invocation syntax: whitespace after `/L` is totally wrong.

Also capitalized `/l` for aesthetic reason: to clearly distinguish from capital I.
Also removed line breaks so that the command could be copy-pasted more safely.
See: https://msdn.microsoft.com/en-us/library/windows/desktop/aa367988%28v=vs.85%29.aspx


lizap:
Thanks, Valentin - we'll take a look at your fixes.

lizap:
Ok, we've reviewed them and are merging now. You'll see the changes live the next time we publish the library. Thanks!!

PR #237, opened 2017-10-19
Added note on async stretch failover

nedpyle:
Behavior change in RS3


lizap:
Hi, Ned - same comment here on branding, plus you have "THis" instead of "This" in the first word of one of your paras. Thanks, LP

nedpyle:
Done thanks

NP | PM

PR #243, opened 2017-10-23
Correcting errors

mitiih:
Line 43 about Exchange ActiveSync pre-authentication, corrected information

Line 66-69, this looks incorrect. Fixed as it should be enabled if URLs are different. Refer to: https://technet.microsoft.com/en-us/library/dn383995(v=ws.11).aspx

Lines 104-107, same as 66-69.


lizap:
Hi - thanks for the contribution. I'm going to have one of our team review the changes and will come back with any comments. One request, though - please remove /en-us/ from the link. We don't include locale information - that way if a localized version is available, and the user's locale is set, we can render the article in the same locale. Thanks!

PR #245, opened 2017-10-24
Links to Microsoft sites should use HTTPS instead of HTTP

DuncanmaMSFT:
Also fixed some odd wording and typos in one file (separate commit if look at the PR)


lizap:
Thanks for the fix, Duncan!

PR #249, opened 2017-10-27
Shrink down circle image

pwiglemsft:
Can we shrink down the circle image? It's super large, too large imho.


lizap:
Hi, Pieter - that method of resizing the art doesn't work with our tools (oh, how I tried), but this did: Always On VPN Infrastructure 

Do you want to replace the markdown image reference with an HTML image reference and then resubmit?

Thanks!

lizap:
Hi, Pieter- I've discussed your suggestion regarding the size of the graphic with the author of the article. He points out that when we size down the graphic, the font is tough to read. We'll look at redesigning the graphic to balance legibility with size/whitespace.

Thanks for raising the issue. For now, I'm closing this pull request without merging. Thanks - Liza

PR #255, opened 2017-11-01
Switch to more accurate title

richlander:
The current title is not descriptive of the content, but (conceptually) one third of it.


lizap:
Thanks, Rich - I've passed the suggested change along.

lizap:
Hi, Rich - I've passed along your input to the writer, who'll take it under advisement for the next rewrite. For now, he'd prefer to leave it as is, since the SAC is the new issue that we need customers to discover/understand. Thanks!

PR #257, opened 2017-11-05
Fix code formatting

psignoret:


lizap:
Good catch - thanks, Phillipe! Merging now.

PR #258, opened 2017-11-05
Correct system requirements

MikeBazMSFT:
The original requirements documentation includes several nonsensical statements that would technically mean Azure or Generation 1 virtual machines are not supported.  This includes PXE and no-IDE.  It also includes an inappropriate reference to Nano early on.

### DIFF
--- a/WindowsServerDocs/get-started/System-Requirements.md
+++ b/WindowsServerDocs/get-started/System-Requirements.md
@@ -26,7 +26,7 @@ This topic addresses the minimum system requirements to run Windows Server&reg; 
 >   
 
 > [!NOTE]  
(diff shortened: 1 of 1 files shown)


lizap:
Thanks, Mike - I've passed this along to one of the writers and will see about incorporating as soon as possible.

jaimeo:
Sorry for the delay--we've been doing extensive review of the existing system requirements doc and the proposed changes with the internal SMEs for this area. They have confirmed and re-asserted that the existing published version is absolutely correct and accurate and have advised no changes to this topic at all at this time. If there are different requirement details for Azure systems, then we should look at creating a separate system requirements topic in the Azure documentation area that clarifies those differences.

lizap:
Closing this PR as outside the scope of this tooling/this process. If you have concerns, contact elizapo.

PR #262, opened 2017-11-09
Fix missing markdown escapes in shutdown.md

rico-chet:
Description for the /d parameter was broken due to unescaped characters which were interpreted by markdown.


lizap:
Thanks for fixing that! Merging now, so you should see the change this morning. Thanks, Liza

PR #263, opened 2017-11-13
Added short section on VM guest state

larsiwer:


lizap:
Thanks, Lars! You should see this addition live later this afternoon. Thanks, Liza

PR #269, opened 2017-11-20
Update guarded-fabric-install-trusted-tpm-root-certificates.md

aseigler:
Link to TrustedTPM.cab is HTTP, not HTTPS.


lizap:
Thanks, Alex - looks like a batch change to other links that are supposed to be "https" hit this file erroneously. Thanks for fixing - you'll see this change live on Tuesday, 11/21.

PR #278, opened 2017-11-28
Fix image references

jongalloway:
Converted image references from img tags to markdown syntax. Previous version looks fine on GitHub but isn't rendered on docs.microsoft.com - just shows escaped HTML.


lizap:
Thanks, Jon! You should see your fixes later this afternoon.

PR #290, opened 2017-11-30
Updated software requirement

trahul87:
Updated the software requirement for Windows Server 2016.


lizap:
Thanks for the addition! You should see it live later today.

PR #301, opened 2017-12-09
Update Forest-Wide-Updates.md

Thameur-BOURBITA:
I think that's still important to mention in the schema update documentation that rollback is only possible through a entire restoration of the forest.
Regarding the automatic preparation of the forest during the installation of the first domain controller, it is important to indicate that with a account membre of Domain Admin group , the preparation of the forest will fail, it needs other privileges at the forest level .

### DIFF
--- a/WindowsServerDocs/identity/ad-ds/deploy/RODC/Forest-Wide-Updates.md
+++ b/WindowsServerDocs/identity/ad-ds/deploy/RODC/Forest-Wide-Updates.md
@@ -18,7 +18,9 @@ ms.technology: identity-adds
 You can review the following set of changes to help understand and prepare for the schema updates that are performed by adprep /forestprep in Windows Server 2012 R2 and Windows Server 2012.
 
 The schema update is irreversible ,for the rollback , you have restore the entire forest.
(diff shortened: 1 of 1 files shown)


Thameur-BOURBITA:
Concerning the link https://technet.microsoft.com/library/dd464018(v=ws.10).aspx , it doesn't talk about Windows 2012 ,Windows 2012 R2 and Windows 2016, I suggest to replace it by this one : https://docs.microsoft.com/en-us/windows-server/identity/ad-ds/deploy/what-s-new-in-active-directory-domain-services-installation-and-removal

lizap:
Thanks, Thameur - I'm sending this over to the AD writing team to take a look. Thanks, Liza

MicrosoftGuyJFlo:
@Thameur-BOURBITA Thanks for your contribution but this article is not about adprep. While it is important to know what happens when something goes wrong this is not the article to do that in. 

I will keep this info in mind as we work on updating the rest of the AD content so that it is brought out where it needs to be.

MicrosoftGuyJFlo:
#please-close

PR #315, opened 2017-12-21
Update refs-overview.md

mdragan:
Data Deduplication is not available using ReFS.


lizap:
Hi, Mateusz - I spoke with the engineering team to understand what's going on in this article. Data deduplication IS supported but there are some considerations that we need to better document. We're working on updating the table with additional information. As such I'm going to close this PR for now. Thanks for raising the issue.

mdragan:
Thanks Liza.
Better documented always welcome.

PR #325, opened 2018-01-05
Remove Workplace-Joined Device Requirement

AdamDunn:
Now that Hotfix 3035025 [ https://support.microsoft.com/en-us/help/3035025/hotfix-for-update-password-feature-so-that-users-are-not-required-to-u ] has been released, it is no longer required that devices be registered to the workplace in order to use the password reset page.


lizap:
Thanks, Adam! You should see this change live later today.

PR #330, opened 2018-01-09
New cluster-performance-history.md topic v1

cosmosdarwin:
Please don't add to the T-o-C just yet, this is for Insiders. :-)


lizap:
Hi, Cosmos - merging now. When you add this to the TOC, though, can you also add a description to the metadata? That will help with discoverability in the future.

This article should be live by 4 PM on 1/9/18.

Thanks!

PR #334, opened 2018-01-10
I update the article based on CI73324 which is requested by Herbert

Deland-Han:
see https://vstf-us-wa-28.partners.extranet.microsoft.com:8443/tfs/ContentIdea/ContentIdea/_workitems?_a=edit&id=73324&triage=true


Deland-Han:
sign-off

Deland-Han:
#sign-off

PR #337, opened 2018-01-12
Fixed the link for Set-NetIPInterface to point to proper MS docs location

kirkdmo:
Updated link for Set-NetIPInterface to point to the proper docs location


lizap:
Thanks for fixing this!

PR #342, opened 2018-01-17
 First take at Symmetry considerations for Storage Spaces Direct (new topic)

cosmosdarwin:
I'll coordinate with Jason to add to ToC when ready.


WindowsServerDocs/storage/storage-spaces/drive-symmetry-considerations.md
```diff
@@ -0,0 +1,175 @@
...
+All servers must have the same number of drives of each type.
+
+For example, if one server has *six* HDD, they must all have *six* HDD.
+
+   > [!NOTE]
+   > It is okay for the number of drives to differ temporarily during failures or while adding or removing drives.
+
+### Model
+
+It is recommended to use drives of the same model and firmware version whenever possible. If you can't, carefully select drives which are as similar as possible and verify that they don't have any conflicting requirements, such as system, adapter, or driver incompatibility.
```
JasonGerend: I recommend using "We recommend" - makes it active voice. Also "which are" [should be](www.quickanddirtytips.com/education/grammar/which-versus-that-0) "that are" in this case.

WindowsServerDocs/storage/storage-spaces/drive-symmetry-considerations.md
```diff
@@ -0,0 +1,175 @@
...
+
+It is not advisable to mix-and-match drives of the same type with sharply different performance or endurance characteristics (unless one is cache and the other is capacity) because Storage Spaces Direct distributes IO evenly and does not discriminate based on model.
```
JasonGerend: (Global) MS voice guidelines recommend using contractions to make the tone more conversational. But would probably be better to rephrase in active voice "Don't mix and match..." or "We recommend not mixing and matching...". Do you want to say something like "...based on model, so the performance of all drives of a type will be lowered to the performance of the slowest drive of that type."

WindowsServerDocs/storage/storage-spaces/drive-symmetry-considerations.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Size
+
+It is recommended to use drives of the same sizes whenever possible.
```
JasonGerend: Another instance of passive voice - I recommend saying "We recommend..." or just "Use drives of the same sizes..."

WindowsServerDocs/storage/storage-spaces/drive-symmetry-considerations.md
```diff
@@ -0,0 +1,175 @@
...
+
+Using cache drives of different sizes may not improve cache performance – see the next section.
+
+Using capacity drives of different sizes may result in stranded capacity – see the next section.
```
JasonGerend: I'm worried that readers might not understand "stranded capacity" - maybe instead of say "might result in some unusable capacity". Or "...result in some stranded (unusable)" if you're sold on "stranded"? I do think it's more clear to say "unusable" since I think readers will have to translate "stranded" into "unusable" in their heads. I get the nuance you're trying to make here - it's not globally unusable - it's just stranded waiting for a rescue. I do see the advantage of the softer phrasing, so I'll defer to you on how you want to handle this. :)

WindowsServerDocs/storage/storage-spaces/drive-symmetry-considerations.md
```diff
@@ -0,0 +1,175 @@
...
+
+Server 1 has NVMe but the others don't.
+
+| Server 1             | Server 2            | Server 3            |
+|----------------------|---------------------|---------------------|
+| 6 x <mark>NVMe</mark> (cache) | -                   | -                   |
+| -                    | 6 x SSD (cache)     | 6 x SSD (cache)     |
+| 18 x HDD (capacity)  | 18 x HDD (capacity) | 18 x HDD (capacity) |
+
+This is not supported. The types of drives must be the same in every server.
```
JasonGerend: As mentioned before, it'd be good to do a search and replace to put contractions in for "is not" "that is", etc.

PR #343, opened 2018-01-17
Removed "Applied to : Windows Server Essentials 2016"

RomanNG:
The Essential Media Pack software is not compatible with 2016 version of Windows Server Essentials. When we try to install, it throw an error : "The Windows Server Essentials Media Pack cannot be installed on this version of the operating system".


lizap:
Hi, Roman - thanks for the suggestion. We're just going to vet it quickly then will get it merged.

Thanks!

PR #357, opened 2018-01-25
ARM support is now GA for AAD-DS

mahesh-unnikrishnan:


lizap:
Thanks, Mahesh! You'll see the change live on Friday morning.

PR #381, opened 2018-02-06
Remove duplicate sentence from first !Note section.

RussoB:
The line "Additionally, DRS requires that you have at least one global catalog server in your forest root domain." is stated in succession.


lizap:
Great catch, thanks!

PR #384, opened 2018-02-08
Grammar fixes in RDP Restrictions

djflux:
There was a missing space in the RDP Restrictions section. Also changed the "that is, jump servers" phrase to an e.g..


lizap:
Thanks for the fix. I made one change - we don't use "e.g" so I changed it to "for example."

Thanks! You'll see the change live later on Friday, 2/9.

PR #409, opened 2018-02-20
Improve AD FS 2016 / MFA link

MikeBazMSFT:
The original link for "Configure AD FS 2016 and Azure MFA" did not actually go to that page; it does now.


lizap:
Thanks, Mike! You should see the fix live later this afternoon (2/20).

PR #415, opened 2018-02-23
Fix links for the "Create a Relying Party Trust" section

MikeBazMSFT:
Added the missing bookmarks to the "Create a Relying Party Trust" section and fix those links to be absolute.  Other links were not touched, so they don't necessarily follow the current recommendations, but changing those was out of scope for this fix.


lizap:
Hi, Mike - can you help me understand the reason for changing the relative links to absolute? I get adding the bookmark targets, but by changing the links from relative to absolute, our build tools can no longer automatically validate the links. As long as the links are in the same docset, which these are, we use relative links for that validation. Thanks, Liza

MikeBazMSFT:
I can look later, but on another edit I did, I was told relative working should be considered an accident, and absolute was the right answer. I will forward later when not travelling.

Sent from my Windows 10 phone - I work for MS and I use our products

________________________________

lizap:
Thanks, Mike. As a rule, though, if you’re linking to a topic in the same content set, you should use a relative link. You can identify the content set by looking at what comes after “docs.microsoft.com” in the URL. For example, the topic you’re editing is https://docs.microsoft.com/en-us/windows-server/identity/ad-fs/deployment/Checklist--Configuring-the-Resource-Partner-Organization. In this case, the content set = windows-server. The article you want to link to is https://docs.microsoft.com/en-us/windows-server/identity/ad-fs/operations/create-a-relying-party-trust. Still windows-server. Can you please change those links back to relative links? Then we can merge the bookmark targets you added. Thanks!

MikeBazMSFT:
ok I think this is good to go now

lizap:
Thanks, Mike! Merging now.

PR #418, opened 2018-02-24
Remove duplicated section (fixes #319)

### LINKED ISSUE #319: "Are third party proxies supported with AD FS?"
The article contains the section "Are third party proxies supported with AD FS?" twice.

psychonic:
As noted in #319, the "Are third party proxies supported with AD FS?" FAQ entry exists twice (under both Deployment and Design). The duplicate was added in 6892d0ccd6ffcbef2f3de30f4a628b719d6cc7d1. Removing that one as the entry seems to be better suited with the other Design questions.


lizap:
Thanks!

Issue #300, opened 2017-12-06
AD FS and HSTS


lol768:
The FAQ states the following with respect to "HTTP String [sic] Transport Security" support:

>HTTP Strict Transport Security (HSTS) is a web security policy mechanism which helps mitigate protocol downgrade attacks and cookie hijacking for services that have both HTTP and HTTPS endpoints. It allows web servers to declare that web browsers (or other complying user agents) should only interact with it using HTTPS and never via the HTTP protocol.

> All AD FS endpoints for web authentication traffic are opened exclusively over HTTPS. As a result, AD FS effectively mitigates the threats that HTTP String Transport Security policy mechanism provides (by design there is no downgrade to HTTP since there are no listeners in HTTP). In addition, AD FS prevents the cookies from being sent to another server with HTTP protocol endpoints by marking all cookies with the secure flag.

> Therefore, implementing HSTS on an AD FS server is not required because it can never be downgraded. For compliance purposes, AD FS servers meet these requirements because they can never use HTTP and all cookies are marked secure.

I disagree with this answer because I feel that it neglects to consider an attacker with active MitM capabilities - where AD FS _not listening on port 80_ is irrelevant (and it's probably a bad idea [anyway](https://scotthelme.co.uk/why-closing-port-80-is-bad-for-security/)). Consider the following scenario:

* Attacker is able to intercept and respond to traffic on the network.
* Attacker uses something like `sslstrip` to coerce the victim's browser into requesting the AD FS URL in plaintext (maybe the victim visits an application over `http:` which redirects to AD FS for auth).
* Attacker intercepts the response and switches the redirect to use `http`. There's no HSTS so the browser happily tries to connect using port 80.
* Attacker proxies requests from AD FS using port 443, stripping Secure flag on any `Set-Cookie` headers.
* Credentials are captured.

Is there anything I have missed which would mitigate this sort of attack?

daveba:
We will close this issue now.  Please reopen if you would like additional follow-up.

lol768:
>We will close this issue now. Please reopen if you would like additional follow-up.

Issue has not been addressed, nor can I re-open a closed issue myself. I don't see why you've closed it.

The page here https://docs.microsoft.com/en-us/windows-server/identity/ad-fs/operations/customize-http-security-headers-ad-fs _still_ misses the point and erroneously states:

[...]

PR #419, opened 2018-02-24
Fix typo.

psychonic:
Fix typo mentioned in #300 (but not the actual focus of that issue), "HTTP String Transport Security" -> "HTTP Strict Transport Security"


lizap:
Thanks!

PR #420, opened 2018-02-26
Removing old references to server 2008

davegarner:


lizap:
Thanks, Dave! I made one change to your submission - in the first occurrence, let's just say "Windows Server" instead of listing the versions. If we were to list the versions, we'd have to list them each fully, so it'd be "Windows Server (Semi-Annual Channel), Windows Server 2016, Windows Server 2012R2, and Windows Server 2012."

PR #427, opened 2018-03-01
Specifying Deduplication support on 1709 & later.

Anhansen:


JasonGerend:
Looks good, though we should double-check once published to make sure it renders properly. Ready to merge. Thanks!

PR #429, opened 2018-03-02
Update features are available on ReFS table

kaushika-msft:
The table mentions that Dedup can be Enabled on the ReFS volume. However, that is not true for Windows Server 2016 (RS1), Windows Server 2012 R2, Windows Server 2012.


lizap:
Hi, Kaushik - can you check the change made by a PM to this same topic? I think it may address the issue: https://github.com/MicrosoftDocs/windowsserverdocs/pull/427

Thanks!

kaushika-msft:
Hi Liza, the change made by PM looks good. Do ensure to correct the overlapping text:
https://github.com/MicrosoftDocs/windowsserverdocs/pull/439

lizap:
Thanks, Kaushik - will do. This merge takes a bit more work, because of that overlap, but I should have it published out this afternoon (Friday, 3/9).

PR #431, opened 2018-03-02
Update hyperv-network-virtualization-technical-details-windows-server.md

adelina2712:
There is no documentation regarding the incompatibiliyt of Set switch and hnv v1 and we had customers opening cases with this issue and complaining afterwards that this is not documented


lizap:
@adelina2712 - Thanks for the addition! I'm merging now.

PR #436, opened 2018-03-05
Defrag: Update Applies To tag for client SKU

ChandKV:
Update Applies To tag to include Windows 10 as
we have defrag on clients too.


lizap:
Thanks!

PR #437, opened 2018-03-07
update group managed service accounts

jischell-msft:
correcting when Group Managed Service Accounts were available - was written as 2008r2, should be 2012


lizap:
Thanks for the fix!

PR #438, opened 2018-03-07
Update server-core-administer.md

m4k9:
Added "=" to "name" parameter in "netsh interface ipv4 set address" command with regard to instruction step "Change to a static IP address" in "Administrative tasks from the command line"


m4k9:
Found more corrections to make. Closing pull request for now.

PR #440, opened 2018-03-07
Add question for refresh token issuance

anandyadavMSFT:
Adding question of refresh token issuance?


anandyadavMSFT:
@billmath can we please pull this change to master? Please check the language of the answer added before merging.

billmath:
#sign-off

PR #446, opened 2018-03-08
escape vertical bar so markdown table parses

rarrarrarr:
Currently, the markdown table parses like so:

image


lizap:
Good catch, thanks!!

PR #447, opened 2018-03-08
Update AD-Forest-Recovery-Single-Domain-in-Multidomain-Recovery.md

UlfBSW:
As part of the #MVPHackaDoc, I found some (hopefully) valuable remarks to clarify timing of unhost/rehost, also ensure that the administrator account is at the beginning and the built-in admin wouldn't help since he cannot logon to a domain controller with ADDS running. The Domain Administrator is needed. In further edits, examples for unhost and rehost might be valuable, since you need to sprecify the partition which need to be unhosted.


lizap:
Thanks, Ulf! You'll see your changes live on 3/9.

PR #449, opened 2018-03-08
MVPHackADoc

mihapihler:
Changed Sign in to Sign-in


lizap:
Thanks for the fix, perhaps the first of many? (Not that we want to have many, many problems needing fixes.) I did change the first "sign-in" back to "sign in" ("Sign in with Windows Hello...") - in that instance, "sign in" is a verb (so 2 words). The second change you proposed is correct as it's used as a noun (sign-in). Thanks!

PR #454, opened 2018-03-12
Misleading which edition ID to be used

LokiMidgard:
The documentation told me to use Get-CurrentEdition and note it, than call Set-Edition with edition ID.

But you need to call set-edition with the edition you want to upgrade to, not the current one. You can get the target edition by calling `/Get-TargetEditions`

In additon the output of the command Get-CurrentEdition is:
```
Deployment Image Servicing and Management tool
Version: 10.0.14393.0

Image Version: 10.0.14393.0

Current edition is:

Current Edition : ServerStandard
```
There is no entry edition ID. I guessed that edition ID is ServerStandard, which turned out to be true, but it felt disquiting. So I added a sample what the ID of the standard edition looks lile.


lizap:
Thanks for adding the clarification!

PR #457, opened 2018-03-13
Correcting a typo in the page

kabala123:
Error results are returned when a role does not satisfy the conditions of a best practice rule, and functionality problems can be expected.

The word "role" must be replaced with "rule" as per the context of the sentence.


kabala123:
This change is not required.

PR #477, opened 2018-03-16
Defrag: Add scheduled task's policy about SSD

ChandKV:
Include the hardcoded policy choices, with respect
to SSDs,  when running  defrag  from the scheduled
task.


lizap:
@JasonGerend - Jason, can you review and let me know if I can merge? Thanks, LP

ChandKV:
@JasonGerend / @lizap, if we can get a page for **DfrgUI** and include this in that page, it would be even better. But till then, I thought it's good to have it at least here as we can redirect customers if they have questions regarding the scheduled tasks policy.

PR #484, opened 2018-03-16
Update vm-load-balancing-overview.md

EldenChristensen:
Added embedded overview video


lizap:
Hi, Elden - due to privacy requirements, we need to use the [!VIDEO] extension, instead of <iframe>. Can you change this? Let me know if you have questions.

PR #495, opened 2018-03-19
Update virtualization.md

k-takai:
Change reference from MSDN to Docs.


lizap:
Thanks for the change!

PR #519, opened 2018-03-26
Update Dual-parity efficiency

DarrylvanderPeijl:
The "8 servers: 80%" is wrong as it takes 16 servers to achieve 80%, even with all-flash.
Either we change the "8 servers" to "16 servers" OR change "80%" to "66.7%"


JasonGerend:
Great catch! This was my mistake - thanks for noticing and making the fix. 
This is ready to publish.

PR #538, opened 2018-03-28
Fix a typo

aaroneg:
I can't imagine any circumstance where this would be ok to disable.


lizap:
Great catch, thanks! You'll see your change live on 3/29.

PR #634, opened 2018-04-13
Correct sqlcmd commands for WID DB detach

aydeisen:
* Removed extraneous backslashes preceding (escaping?) the hash
* put `use master` on its own line, followed by another `GO`
* combined in-line code blocks into single block using code fencing, and defined language as 'batchfile'


lizap:
Thanks for the fix! You should see it live by 4/16/2018.

PR #654, opened 2018-04-18
Update Configure-AD-FS-to-authenticate-users-stored-in-LDAP-directori…

maweeras-msft:
…es.md

This method works for non Microsoft LDAP instances. This feedback was also shared with the PM internally some time back.


lizap:
@billmath and @MicrosoftGuyJFlo  - Bill and John - can you review these proposed changes and let me know if ok to merge? Thanks!

billmath:
@lizap let me look at this.  Logically I understand what @maweeras-msft is proposing but I am not sure we want to recommend "storing" the password.  Especially if they use the PowerShell to create a script which can be opened with notepad.

lizap:
Thanks, Bill - will do.

@maweeras-msft - thanks for raising this issue. Bill's looking into it. While he sorts that, I'm going to temporarily close your PR. Let me know if you have any concerns. Thanks!

Issue #623, opened 2018-04-13
Wrong Image in ADFS oidc Docs


HendrikFW:
Hey, 
i just read the [docs ](https://docs.microsoft.com/en-us/windows-server/identity/ad-fs/development/enabling-openid-connect-with-ad-fs) about to how build an web application using Open ID Connect on ADFS and there is a wrong screenshot at "*11. On the Configure Web API screen, enter the following..*".   
This Image belongs to "*13. On the Configure Application Permissions screen*". The image for step 11 is missing.

illfated:
Resolved in commit https://github.com/MicrosoftDocs/windowsserverdocs/commit/4ae84fe54041184ad3006ad369e785dd84d5e8cf ( PR #662 **Update enabling-openId-connect-with-ad-fs.md** ).

But the page has also changed a lot since that correction, so this ticket is definitely outdated.

illfated:
#please-close

PR #662, opened 2018-04-19
Update enabling-openId-connect-with-ad-fs.md

joel-g:
fixed order of steps 11 and 12


lizap:
Good catch, thanks!

illfated:
(ref. #623)

PR #680, opened 2018-04-23
Updating two new regkeys to turn off USN Journal for the entier serve…

sophyw:
…r or for only some but not all volumes


lizap:
Closing until closer to 1803 ship. Will publish then. Thanks!

PR #718, opened 2018-04-28
Found 2 Broken Links

akashkanaujiya:
Here, I found 2 Broken Links, So I Though, I should Remove these Links. In change, We added 3 Unique Content Links that Related From Windows 10. Where I told People, How to Reset Windows 10 password with Bootable USB Drive, and Also Protect Bootable USB From Virus and How to Make Bootable USB for other OS such as Kali Linux.


lizap:
@akashkanaujiya - thanks for finding the broken links! The target articles are actually in the same library as this, so I need to update to point to those, as opposed to the suggestions you've made. I've done that on my end - you should see fixed links on Monday, 5/7. Thanks!

PR #720, opened 2018-04-29
Add qualifier for documented location of ad://ext/AuthenticationSilo

GSam:
All the other objects listed mention this fact (e.g. create in the Configuration partition), so this just makes it consistent.


MicrosoftGuyJFlo:
#sign-off

PR #745, opened 2018-05-03
Fix formatting for at command

wanion:
Lots of missing line breaks in the Markdown and other weirdness. Should now be readable.

- "Remarks" heading head capitalised (instead of "remarks")
- Syntax now uses `code` formatting like other documentation entries
- Parameters uses a table like other documentation entries which fixes missing line breaks as well
- Remarks and Examples now have line breaks correctly included and use `code` formatting for examples


lizap:
Thanks for fixing this!

PR #749, opened 2018-05-03
escaping curlies and pipes

csharmath:
so they show up as literals in the parameters table.
Probably this is not the only one, just saw in the published PDF and wanted to see why isn't it rendered properly.


csharmath:
Please review my change and just escape the special characters in the docs where it needs to. 
I can't go through a CLA signing procedure with approval request from my company for this right now, so will just close it as is then.

PR #751, opened 2018-05-03
fix echo [on | off] description

UziTech:
fixed the description of `echo [on|off]`

image


UziTech:
looks like this is a problem in multiple places #749

lizap:
Thanks for fixing this!

PR #754, opened 2018-05-04
add known issue for localization

cosmosdarwin:


lizap:
Cosmos - is WAC version 1804? Just curious that it's different from server...

lizap:
Never mind. :) Merging now.

PR #757, opened 2018-05-05
Update bitsadmin.md

doctordns:
Updated list of bitsadmin commands to be in a list instead of a single unreadable paragraph.
Also added Win10 to list of OSs this command is found within.
A minor edit.


lizap:
Thanks for fixing this!

PR #812, opened 2018-05-15
Updated text to help define the sMSA and gMSA acronyms

jasonepperly:
Updated text to define and more consistently use the sMSA and gMSA acronyms


lizap:
Thanks, Jason! I've asked someone over here to do a quick review on your changes. Thanks - Liza

PR #813, opened 2018-05-15
Update Custom-Id-Tokens-in-AD-FS.md

maweeras-msft:
Missing &response_mode=form_post query string parameter to demonstrate allatclaims scope requirements.


lizap:
Thanks!

PR #814, opened 2018-05-15
Update prepare-development-environment.md

eajhnsn1:
Add a note about taking note of search/replace values since they'll be used in later steps.


lizap:
Excellent addition - thanks! You should see the article updated later on 5/15.

PR #817, opened 2018-05-15
Update maintain-servers.md

StevenEk:
Behavior changed for repair.  Incomplete doesn't happen when the node is paused/drained.  Customers noticed, so i changed the documentation on it.


lizap:
Thanks, Steven!

PR #820, opened 2018-05-15
Update performance history docs for Insider build 17666

cosmosdarwin:
- Adopt new series names and cmdlet parameters
- Remove the extra step to print to host
- Remove the troubleshooting (will soon be built into cmdlet)
- Add scripting samples topic and link from main topic


lizap:
Looks good, Cosmos! Merging now, should go live later today.

PR #831, opened 2018-05-16
Update fc.md

doctordns:
Added remark pointing to where fc command is found.

Also, added note to point out it's fc.exe and to be careful since fc is also a built-in alias for Format-Custom.

As an aside, I'd like all the command documents to clarify where this command is found, whether it's still available, and whether it's a builtin command built-into CMD.EXE.


lizap:
Thanks for the added clarification! We'll look at adding the same information to other commands. You'll see this change live on 5/17 (this morning).

PR #833, opened 2018-05-16
Update Use-Remote-Web-Access-in-Windows-Server-Essentials.md

doctordns:
Updated the introduction to explain what RWA *is* and then to explain it's value. 

Trying to improve the page.


lizap:
That's an excellent addition - thank you!

PR #840, opened 2018-05-18
Update enabling-openId-connect-with-ad-fs.md

bspender:
The order of steps 11 and 12 were swapped in regards to the order presented in the wizard being described...


bspender:
thank you!

PR #857, opened 2018-05-22
Update Manage-Hyper-V-integration-services.md

ejsiron:
Mostly expanded information
- original said that "all" Windows integration services are enabled, which isn't quite true
- removed some duplicated verbiage
- reorganized a bit of information to give it better context
- clarified some language, fixed a handful of typos
- mentioned that the host setting could override the guest setting
- added information about newer server SKUs to the guest matrices
- added a link to Sarah Cooley's article on adding/updating guest services in offline VMs


lizap:
Thank you for making these updates!

PR #860, opened 2018-05-22
Update Install-AD-FS-Delegated-Admin.md

JarrettRenshaw:
Approved in CSS content triage as Content Idea Request 83375:Update: Add note with older PS value to Creating an AD FS Farm without domain admin privileges. Reviewed by Ryan Ries.


lizap:
Thanks, Jarrett!

PR #884, opened 2018-05-27
Fix broken link to "Integrated Windows Authentication"

justink-ase:
This change fixes a broken link in the documentation


lizap:
Thanks for the fix! You'll see the change live on 5/29.

PR #910, opened 2018-05-31
Change WSUS using URLs list

Rei-Ikei:
The following information appears to be correct.
I think that HTTP and HTTPs are changed in some URLs unexpectedly.

https://web.archive.org/web/20170710190910/https://technet.microsoft.com/en-us/library/hh852346.aspx

•	http://windowsupdate.microsoft.com
•	http://*.windowsupdate.microsoft.com
•	https://*.windowsupdate.microsoft.com
•	http://*.update.microsoft.com
•	https://*.update.microsoft.com
•	http://*.windowsupdate.com
•	http://download.windowsupdate.com
•	http://download.microsoft.com
•	http://*.download.windowsupdate.com
•	http://wustat.windows.com
•	http://ntservicepack.microsoft.com
•	http://go.microsoft.com

### DIFF
--- a/WindowsServerDocs/administration/windows-server-update-services/deploy/2-configure-wsus.md
+++ b/WindowsServerDocs/administration/windows-server-update-services/deploy/2-configure-wsus.md
@@ -52,11 +52,11 @@ if there is a corporate firewall between WSUS and the Internet, you might have t
 
 -   https://windowsupdate.microsoft.com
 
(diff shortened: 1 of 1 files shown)


lizap:
Thank you for fixing these! It looks like a batch update to change actual live links to the more secure https:// mistakenly changed these as well. You'll see the fix live on 5/31.

PR #931, opened 2018-06-05
Rename New-KdsRootKey to Add-KdsRootKey

Tadas:
There is no `New-KdsRootKey` cmdlet, it should be `Add-KdsRootKey`

List of KDS module cmdlets: https://docs.microsoft.com/en-us/powershell/module/kds/?view=win10-ps


lizap:
Thanks for fixing this!

PR #934, opened 2018-06-05
Update nps-top.md

yoosi:
Modify a hyperlink to make the visible text more verbose so it matches the convention used on the following line.


lizap:
Thanks!

PR #935, opened 2018-06-05
Fixed parameter name for template disk type

rpsqrd:


lizap:
Thanks! You'll see the fix live on 6/6.

Issue #962, opened 2018-06-10
Broken docs for attrib's parameters


AviKav:
The section of the parameter table describing parameters that alter file attributes is rendered incorrectly due to unescaped pipes.

Current table:

Parameter | Description
------------ | -------------
{+ | -}r
{+ | -}a
{+ | -}s
{+ | -}h
{+ | -}i
... | ...

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: ce055b59-7d93-7c0c-6373-109b53f31119
* Version Independent ID: fc5ac67c-1db4-bfca-a71a-bb1f2c4296df
* Content: [attrib](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/attrib)
* Content Source: [WindowsServerDocs/administration/windows-commands/attrib.md](https://github.com/MicrosoftDocs/windowsserverdocs/blob/master/WindowsServerDocs/administration/windows-commands/attrib.md)
* Product: **windows-server-threshold**
* GitHub Login: @coreyp-at-msft
* Microsoft Alias: **coreyp**

lizap:
Hey, thanks for fixing this ugly table! Closing out the issue with many, many thanks.

PR #963, opened 2018-06-10
Fix attrib's Parameters table

AviKav:
Github issue:  #962


lizap:
Thanks so much for fixing this! Obviously our HTML conversion to markdown went sideways. Thanks, Liza

PR #967, opened 2018-06-11
Fixed table in schtasks.md

SteGriff:
Fixed a markdown table with no heading which is not displaying correctly on docs.microsoft.com


lizap:
Thanks for catching this table error! The conversion from HTML to markdown has a few glitches, apparently - thanks for fixing it. - Liza

PR #980, opened 2018-06-14
Update KMS client setup keys page

kaushika-msft:
Added Windows Server, version 1803 client keys. Please check and update the client keys for Windows 10, version 1803.


lizap:
@jaimeo - Do you have the client keys so we can add those prior to publishing this to the server library?

lizap:
Merging now - will add client keys in private.

Issue #693, opened 2018-04-25
Skype for Business Server 2015 Compatability


aakrol:
The website that is referenced for skype for business server 2015 - https://docs.microsoft.com/en-au/skypeforbusiness/plan-your-deployment/requirements-for-your-environment/server-requirements#operating-systems-for-skype-for-business-server-2015 reflects that it is supported on Server 2016 now. 

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 490f6e58-73b1-52d8-521f-54a60c5425b0
* Version Independent ID: e5ea308f-7ff5-352d-0ba5-4f384e0186b8
* Content: [Microsoft Server Application Compatibility](https://docs.microsoft.com/en-us/windows-server/get-started/server-application-compatibility)
* Content Source: [WindowsServerDocs/get-started/Server-Application-Compatibility.md](https://github.com/MicrosoftDocs/windowsserverdocs/blob/master/WindowsServerDocs/get-started/Server-Application-Compatibility.md)
* Service: **unspecified**
* GitHub Login: @jaimeo
* Microsoft Alias: **jaimeo**

illfated:
Ref. commits https://github.com/MicrosoftDocs/windowsserverdocs/commit/bc0a54fecbe2aef8b4ac4232352e1b09b60009f8 & https://github.com/MicrosoftDocs/windowsserverdocs/commit/8cda1fa0625dec77bf1476a7424660672bfaa290 from PR #990 .

illfated:
#please-close

PR #990, opened 2018-06-16
From private repo

huypub:


illfated:
(ref. "Skype for Business Server 2015 Compatability" #693)

PR #992, opened 2018-06-16
Fix up /remove entry for icacls.

Ionic:
Markdown misinterpreted the pipe character as a table column delimiter,
so the actual description was never shown.

Also, while at it, add a newline at EOF as a courtesy.


lizap:
Thanks!

PR #1002, opened 2018-06-19
Update Install-the-Federation-Service-Proxy-Role-Service.md

idMdev:
Hi, the documentation is incorrect and we have a customer comment that points that out. Proposing changes to accurately reflect the installation steps for the Web Application Proxy role for AD FS


lizap:
Fantastic! Thanks for fixing this. You'll see the changes live today (6/19).

PR #1003, opened 2018-06-19
Fix the MinRoot link

DavidBerg-MSFT:
Fix the link to MinRoot, it's going to the wrong page.  (It's going to the general requirements page, https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/reference/hyper-v-requirements, which says nothing about min root, it should be going to https://docs.microsoft.com/en-us/windows-server/virtualization/hyper-v/manage/manage-hyper-v-minroot-2016, like it says in the link).


lizap:
Thanks for fixing this! You'll see the change live today (6/19).

PR #1011, opened 2018-06-20
Update installation-options.md

doctordns:
Add a note to the effect that Admin Center is non-operational using the default browser (IE). Whilst this needs to get fixed in due course, in the meantime a note to let users know that IE doe\s NOT work with AC.


lizap:
Thanks for adding this clarification! I'm going to run the language through the WAC team and will reply back here with any changes. Thanks, Liza

doctordns:
That's fine Lisa. If you or they can improve the language, then please do so.

lizap:
Hi - the team have updated this article with the following information:

Important
You can't use Internet Explorer to manage Windows Admin Center - instead you need to use a supported browser. If you are installing Windows Admin Center on Windows Server, we recommend managing by connecting remotely with Windows 10 and Edge. Alternatively, you can manage locally on Windows Server if you have installed a supported browser.

Because it addresses the same thing as your PR, I'm going to close out this PR. Thank you SO much for raising the issue!

PR #1031, opened 2018-06-23
Fix up a bad merge

MikeWasson:
Looks like someone left in a conflict marker when resolving a merge conflict


lizap:
Great catch, Mike - thanks@

PR #1033, opened 2018-06-25
Escape parameter choice pipes

CorruptedHeart:
The parameters' choices are separated by pipes `|` which markdown is interpreting as column end tags.

This change escapes those pipes to correctly format the parameter choices and description as part of the table.


lizap:
Thank you!

PR #1068, opened 2018-06-29
fixed Move-VMStorage command

jaromirk:


lizap:
We had another PR that fixed the same problem. As such I'm going to close this one as a duplicate.

PR #1084, opened 2018-07-04
Fix performance-history-scripting.md Example Scripts

comnam90:
Updated example script for `Sample 3: Noisy neighbor? That's write!` as the current implementation didn't work.

 * Moved `Format-Iops` function into the context of the invoke-command so that it could actually be called on the remote machine.    
 * Added an `if` statement to the `do` loop in `Format-Iops` so that if the initial raw value is less than 1000, it doesn't end up being output as `0 K`.    
 * Changed values passed to `Format-Iops` in the `PsCustomObject` block to actually be raw values rather than full `MSCluster_HealthMetric` objects.    
 * Changed `Sort-Object` statement in final output to be descending to actually output the Top 10 rather than the Bottom 10.

### DIFF
--- a/WindowsServerDocs/storage/storage-spaces/performance-history-scripting.md
+++ b/WindowsServerDocs/storage/storage-spaces/performance-history-scripting.md
@@ -220,32 +220,32 @@ The results from every server come together as `$Output`, which we can `Sort-Obj
 Here's the script:
 
 ```
-Function Format-Iops {
-    Param (
-        $RawValue
-    )
-    $i = 0 ; $Labels = ("", "K", "M", "B", "T") # Thousands, millions, billions, trillions...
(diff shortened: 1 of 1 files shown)


comnam90:
Updated example script for `Sample 6: Memory hog, you can run but you can't hide` as the current implementation didn't work.

 * Moved `Format-Bytes` function into the context of the invoke-command so that it could actually be called on the remote machine.    
 * Added an `if` statement to the `do` loop in `Format-Bytes` so that if the initial raw value is less than 1024, it doesn't end up being output as `0 KB`.    
 * Changed values passed to `Format-Bytes` in the `PsCustomObject` block to actually be raw values rather than full `MSCluster_HealthMetric` objects.    
 * Changed `Sort-Object` statement in final output to be descending to actually output the Top 10 rather than the Bottom 10.

cosmosdarwin:
Changes look good to me. Thanks very much, Ben!

Issue #901, opened 2018-05-30
EMET related content should be removed


mahanpathak:
This article still have reference to EMET. Please provide details of implementation with WDEG.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: fd5878a8-a83e-e75d-ddae-59edc04e7138
* Version Independent ID: 0f0cbbe4-20da-3abe-1a7c-effc608d8cca
* Content: [Privileged Access Workstations](https://docs.microsoft.com/en-us/windows-server/identity/securing-privileged-access/privileged-access-workstations)
* Content Source: [WindowsServerDocs/identity/securing-privileged-access/privileged-access-workstations.md](https://github.com/MicrosoftDocs/windowsserverdocs/blob/master/WindowsServerDocs/identity/securing-privileged-access/privileged-access-workstations.md)
* Service: **unspecified**
* Product: **windows-server-threshold**
* GitHub Login: @coreyp-at-msft
* Microsoft Alias: **coreyp**

illfated:
Resolved in PR #1087 (commit https://github.com/MicrosoftDocs/windowsserverdocs/commit/33290f4e051bfcc6b8232fc40da680982295d18b).

illfated:
#please-close

PR #1087, opened 2018-07-04
Terminology, Formatting & Typos

clconlin:
Find & replaced EMET with Windows Defender Exploit Guard - NOTE - We need to add guidance around how exactly to configure this. Could be that the configuration from the latest MS Security Baselines is sufficient.

A number of formatting, grammar and typo fixes.


illfated:
(ref. #901)

PR #1132, opened 2018-07-13
fixing password prompt asterisk location

peterc0der:


lizap:
Thanks for fixing this!

PR #1160, opened 2018-07-16
Replaced the term"disabling" with "locking out"

karammasri:
The second bullet for ESL under the section "Advantages of Extranet Smart Lockout in AD FS 2016" is using the term"disabling". It should be use the term "locking out", as disabling an account has a different meaning to locking out an account.


MicrosoftGuyJFlo:
#sign-off

PR #1204, opened 2018-07-23
Minor formatting touchup

skycommand:
Two code snippets were properly formatted. A stray "2.0" at the bottom which seem to serve no purpose, was deleted. Inline code formatting applied to two executable file names, which had were formatted inconsistently.


lizap:
Thanks for the cleanup! I changed the formatting on command names from monospace font to bold; other than that, I'm merging your changes.

PR #1213, opened 2018-07-24
Styling fix for "wall of logos"

cosmosdarwin:
Greatest hack in human history to work around docs.microsoft.com not allowing CSS class definitions or style tags, and by extension, not supporting CSS selectors like :hover. When there's a will, there's a way. $10 says it works.


lizap:
We'll see... (taking a mental note to also talk about using one of the "new" landing page formats for this page - might be more of what you're trying to get, with less CSS hackage.)

PR #1216, opened 2018-07-24
Added 2016 server to gen1/gen2 table

jaromirk:


lizap:
Thanks for adding this info! You'll see it live Tuesday morning.

PR #1228, opened 2018-07-25
Ditch CSS completely (use JS only) for "wall of logos"

cosmosdarwin:
Yesterday's hack didn't quite cut it, even inline styles (like max-width 225px) get lost in translation. So doubling down with a purely JS-based approach that injects the styles during onload(), onmouseover(), and onmouseout(). To recover my lost bet, $20 says it works this time.


JasonGerend:
I still think you should just resize the image files themselves - it'll make the page load faster as well.

Let's publish this and see what happens.

cosmosdarwin:
As discussed in email - let's try this first. I can resize as last resort.

The images, in their current "retina" (2x the needed resolution) form, are still just 400 KB all together.

lizap:
Merging for publish this afternoon. @cosmosdarwin - if you could check the size around 3:50 today, that'd be great. Thanks!

PR #1229, opened 2018-07-25
updating the incorrect URI for the claim type

maweeras-msft:
The URI for password expiration claims are http. Not https. See https://adfshelp.microsoft.com/AadTrustClaims/GenerateClaims for sample claims rules.


lizap:
Thanks for fixing!

PR #1243, opened 2018-07-26
HTTPS on Claim Uri

davide-rota:
URI for Claim windowsaccountname had HTTPS instead of HTTP


lizap:
Thanks for fixing! We had a bulk change a few months ago that inadvertently changed this. You'll see the fix live today.

PR #1252, opened 2018-07-27
Update plan-hyper-v-security-in-windows-server.md

larsiwer:
Added WS2019 to the list of products this document applies to; tweaked title to be more generic; fixed Installation Options link


coreyp-at-msft:
Hi @larsiwer ,

Thanks for helping out with this. Taking 2016 out of the body is okay, but Microsoft has not released a product called Windows Server 2019, so we can't add that to "Applies to." Pull that reference out and you'll be good to go.

Thanks,
Corey

larsiwer:
Thanks @coreyp-at-msft -- removed WS2019 for now. We'll have to re-add it once WS2019 releases.

PR #1256, opened 2018-07-29
Fix broken link for PowerShell cmdlets

billkindle:
Old link was giving a 404 message, I found the correct link and pasted it in the markdown. This is my very first PR I think.


coreyp-at-msft:
Thanks for the update, @billkindle !

PR #1268, opened 2018-08-01
Removing Server 2016 reference

kaushika-msft:
You cannot install KMS host key of Windows Server 2012 R2 for Windows 10 on Windows Server 2016.


lizap:
Thanks, Kaushik, for the correction. Merging now and will publish today.

PR #1275, opened 2018-08-01
Removed extraneous text

alfredmyers:


lizap:
Thanks for the fix - you'll see the change live today (Friday).

PR #1276, opened 2018-08-01
Correct WDS-Transport entry

MikeBazMSFT:
as described at https://docs.microsoft.com/en-us/windows-server/get-started/whats-new-in-windows-server-1803, 1803 has WDS-Transport


lizap:
Great catch - thanks!

PR #1293, opened 2018-08-03
Fixed the A-Z bookmarks

Welshmilla:


lizap:
Thanks for fixing this! That was a lot of work - we appreciate it!

PR #1295, opened 2018-08-03
Fixed the header for the ksetup:delkdc command

Welshmilla:


lizap:
Thanks for fixing this!

PR #1314, opened 2018-08-08
Corrected number of WID servers from 5 to 30

imav8n:


billmath:
@imav8n Good catch!  Thanks!

billmath:
#sign-off

PR #1326, opened 2018-08-10
Update delimit-volume-allocation.md

DarrylvanderPeijl:
Calculate width does not work when using the script in Powershell ISE as it does not provide a (Get-Host).UI.RawUI.WindowSize.Width property.


JasonGerend:
Looks good - thanks for your contribution! Let's merge this!

PR #1346, opened 2018-08-16
Verify-DNS-Functionality... Fixed the dcdiag syntax + others

drewjanderson:
Updated the dcdiag command to contain the proper characters as well as fixed the position of the `/` so that it would run properly without a syntax error. After making that update I found several other places where commands to be executed contained the markdown formatting that was ignored when putting text between the ` marks.

Honestly just didn't want someone else putting this extra text in the command like I did before I really understood what was going on, so figured I would fix it.


coreyp-at-msft:
Hi @drewjanderson ,

Thanks for taking care of that. We, and future readers of this topic, appreciate it!

Thanks,
Corey

PR #1460, opened 2018-09-08
Readme edits

getex1977:


dawnwood:
@getex1977  what exactly are you trying to accomplish here?

PR #1466, opened 2018-09-10
Fixed some of the formatting

Lampei:
The description was being hidden the way the markup was originally written.


coreyp-at-msft:
Thanks for cleaning that up, @Lampei !

PR #1469, opened 2018-09-10
Updated this doc to include a major caveat about on-prem only customers.

sdabbiru:
This doc does not apply for hybrid customers. I wanted to make it very clear to the users.


coreyp-at-msft:
Hi @sdabbiru , You say "does not apply," but then in the note you say, "only recommended for." Is it that it can't be done in hybrid, or just that it's not recommended? In either case, an explanation of why might be helpful.

Thanks,
Corey

coreyp-at-msft:
Hi @sdabbiru. Are you able to clarify?

Thanks,
Corey

PR #1508, opened 2018-09-20
Update Configuring-Alternate-Login-ID.md

marcelodiiorio:
Fixed format on bulleted item.


billmath:
@marcelodiiorio Good catch!  Thanks!

#sign-off

PR #1512, opened 2018-09-20
Properly escape backslashes and asterisks.

TastesLikeTurkey:


coreyp-at-msft:
Thanks, @TastesLikeTurkey !

PR #1518, opened 2018-09-22
Link repair

skycommand:
1. Repaired a dead link.
2. Replaced an MSDN link that, while being alive, had become completely useless. (I'm sure you'll know what I mean if you [visit it](https://msdn.microsoft.com/library/cc239875.aspx).) I believe the [replacement](https://msdn.microsoft.com/en-us/library/cc239318.aspx) would prove more useful.
3. Included quotation marks in PowerShell snippets to suppress aberrant syntax highlighting. Perhaps the highlighter needs improvements.


JasonGerend:
Looks good! I'll do a little link cleanup after publishing (remove en-us and deprecated, since that has a very specific use in our docs)

coreyp-at-msft:
Thanks, @skycommand !

skycommand:
@JasonGerend While it is not much of a big deal, perhaps I should tell you this anyway: I am aware of the special meaning of "deprecated" within Microsoft; but this document's intended audience are not within Microsoft.

Issue #1566, opened 2018-10-03
404 on Server Folder Setup


peterennis:
Here:

![image](https://user-images.githubusercontent.com/140737/46430547-1f10fd00-c6fe-11e8-8414-d58da9554ffe.png)

peterennis:
https://github.com/MicrosoftDocs/windowsserverdocs/blob/master/WindowsServerDocs/storage/folder-redirection/deploy-folder-redirection.md

PR #1571, opened 2018-10-03
Corrects #1566

exchange12rocks:
The problem affects GitHub preview only


coreyp-at-msft:
thanks, @exchange12rocks !

PR #1581, opened 2018-10-04
WDS is in core in 2019

jaromirk:
In 2019 we added WDS into core edition (Windows Deployment Services and Transport Server)


lizap:
Thanks, Jaromir! Merging now.

PR #1589, opened 2018-10-05
Fix prerequisites table

tksh164:
Fix to the prerequisites table because the prerequisites table was displayed as a code block


coreyp-at-msft:
Thanks, @tksh164 !

PR #1593, opened 2018-10-05
Fix documentation related to Connection Broker under RemoteGuard scenario

vineelko:
Fix documentation related to Connection Broker under RemoteGuard scenario

### DIFF
--- a/WindowsServerDocs/remote/remote-desktop-services/rds-vdi-supported-config.md
+++ b/WindowsServerDocs/remote/remote-desktop-services/rds-vdi-supported-config.md
@@ -36,7 +36,7 @@ The following table outlines which of these new features are supported in a VDI 
 
 Remote Credential Guard is only supported for direct connections to the target machines and not for the ones via Remote Desktop Connection Broker and Remote Desktop Gateway.
 > [!NOTE]
-> Connection Broker which is directly connected(not under DNS round robin or HA) from the client might still work.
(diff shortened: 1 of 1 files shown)


ChristianMontoya:
Hmm...can we change it to something like?
   Remote Credential Guard might still work via Remote Desktop Connection Broker in a single-instance environment and the DNS name matches the computer name.

vineelko:
> Hmm...can we change it to something like?
> Remote Credential Guard might still work via Remote Desktop Connection Broker in a single-instance environment and the DNS name matches the computer name.

Reworded the change to the documentation.

ChristianMontoya:
@vineelko : Sorry, one more change. Can you change it so it says
   where the DNS name matches the computer name.

ChristianMontoya:
@lizap : Content looks good to me now. Can you please review for style?

lizap:
It's a bit contradictory as it is - it's only supported in X scenario, but might work in y scenario? Could we say something like this instead? 

If you have a Connection Broker in a single-instance environment, and the DNS name matches the computer name, you may be able to use Remote Credential Guard, although this is not supported.

ChristianMontoya:
@lizap That works with me.

vineelko:
@ChristianMontoya @lizap Updated it

lizap:
Thanks! Merging now.

PR #1600, opened 2018-10-07
Fix availability date for Windows Server 2016

tksh164:
Ref: https://support.microsoft.com/en-us/lifecycle/search?alpha=Windows%20Server%202016


coreyp-at-msft:
Thnaks, @tksh164 !

PR #1609, opened 2018-10-08
Update deploy-work-folders-adfs-step1.md

vcozier:
execution of powershell script is in two places, removed from command box for setting the execution policy


JasonGerend:
Thanks for making this valuable contribution!!! :)

PR #1651, opened 2018-10-14
fix broken Markdown syntax in headers

cTxplorer:
- Add missing spaces in Markdown headers' syntax
- Add inline code syntax to filenames


coreyp-at-msft:
Thanks, @cTxplorer !

PR #1672, opened 2018-10-14
chore: Fix merge conflict

nschonni:
Came across the merge conflict, and think the intention was to remove the language code from the link so I took that resolution


coreyp-at-msft:
Hi @nschonni, This was fixed in the private repo over the weekend. The fix will be live later today.

Thanks,
Corey

PR #1675, opened 2018-10-14
chore: Fix casing on addgroupsetp images

nschonni:
Gets rid of the second directory seen on the web root because of the case difference of "windowsserverdocs" vs. "WindowsServerDocs"


coreyp-at-msft:
Thanks, @nschonni!

PR #1693, opened 2018-10-15
chore: Remove orphaned copy of troubleshooting-sil

nschonni:
Appears to be an older copy of https://github.com/MicrosoftDocs/windowsserverdocs/blob/master/WindowsServerDocs/administration/software-inventory-logging/troubleshoot-sil.md that ended up in the root


coreyp-at-msft:
Thanks, @nschonni!

PR #1748, opened 2018-10-20
Update ver.md

doctordns:
Updated to note that ver does not work in powershell and provided an example that shows how to get the ver info from PowerShell.


coreyp-at-msft:
Thanks, Thomas!

PR #1749, opened 2018-10-20
Update type.md

doctordns:
Updated to note that type only exists inside cmd.exe/

Also pointed out that type is an alias to get-content (and provided a helpful link.

Just trying to make these pages a bit more modern! ;-)


coreyp-at-msft:
Thanks, Thomas! I tweaked a couple things.

PR #1763, opened 2018-10-23
Update windows-server-release-info.md

k-takai:
Add about Windows Server 2019 and Windows Server, version 1809.


lizap:
Thanks for your work here. We've just updated the [Server release information page](https://docs.microsoft.com/en-us/windows-server/get-started/windows-server-release-info) to reflect the availability of Windows Server 2019 and Windows Server, version 1809. Because we had to change dates, we're not able to use your fix, so I'm going to close this. Thanks.

PR #1783, opened 2018-10-25
Update TargetFirmware BinaryPath in XML example

howardsun1:
TargetFirmware BinaryPath in supported components XML doesn't actually support network paths; the example looks like a network path and might hint to users otherwise.


coreyp-at-msft:
Thanks, @howardsun1 !

PR #1806, opened 2018-10-26
Update README.md

shefaligrg:


coreyp-at-msft:
Hi @shefaligrg, "technical accuracy" is used here in a general sense, and so doesn't require a definitive article to place it in a particular context. We'll go ahead and leave it as is.

PR #1813, opened 2018-10-28
fix: Remove space in markdown links

nschonni:
The space renders these as text instead of links


coreyp-at-msft:
Thanks, @nschonni !

PR #1814, opened 2018-10-28
chore: Remove en-us from azure.microsoft.com links

nschonni:
This matches link locale guidelines


coreyp-at-msft:
Thanks, @nschonni !

PR #1877, opened 2018-11-06
Clean up Configure-Server-Certificate-Autoenrollment.md

JackDDavis:
'Public Key Policies' incorrectly referenced twice. In both line 53 & 54. Only required once.


lizap:
Thanks for catching this!

PR #1898, opened 2018-11-12
Removing known issue

andreiztm:
Removing known issue as the devs reacted very fast to my last follow-up and pushed a fix into prod.


lizap:
Thanks for fixing this!

PR #1916, opened 2018-11-14
Update ad-fs-single-sign-on-settings.md

jozf95:


lizap:
Thanks for fixing this!

PR #1976, opened 2018-11-25
/Z restartable mode

WurstCommander:
more information about the mode, it only works with network devices, not local hard drives. I spent hours trying to get this mode to work
 :-)


lizap:
Thanks for your updates - we'll get this published today (11.27).

PR #2004, opened 2018-11-29
Adding Server 2019 to applies to

kaushika-msft:


lizap:
Thanks, Kaushik - merging now. You should see this change live on 12/10 AM.

PR #2007, opened 2018-11-30
Fixes in scripts

jaromirk:
Get-ClusterNode -> (Get-ClusterNode).Name as invoke command accepts strings (node names)
NetAdapter.Bytes-> NetAdapter.Bandwidth


lizap:
Thanks! You should see the changes live today.

PR #2080, opened 2018-12-13
Clarified use of -Force with -PolicyVersion

rpsqrd:


lizap:
Thanks for the fix - will get this published ASAP.

PR #2083, opened 2018-12-13
Clarify Desktop Experience text

MikeBazMSFT:
The original text ignores the reason for why the named releases don't include the Desktop Experience, and implies that somehow it's broken that that's the case.  It's by design, and considering 1809 and 2019 are being released at the same time, it's important to be clear on the reasoning.

### DIFF
--- a/WindowsServerDocs/get-started-19/whats-new-19.md
+++ b/WindowsServerDocs/get-started-19/whats-new-19.md
@@ -17,7 +17,7 @@ This topic describes some of the new features in Windows Server 2019. Windows Se
 
 ### Desktop experience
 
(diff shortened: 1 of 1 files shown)


lizap:
Hi, Mike - do you think we want to stress that the SAC releases are server core only, and that's why they don't include the Desktop? Or just go with the wording you have now? Thanks!

MikeBazMSFT:
Let me take a look tonight and do another pull request if necessary (since it looks like you committed already 😊)

Sent using Outlook for Android<https://aka.ms/ghei36> - I work for Microsoft, and I use our products

________________________________

MikeBazMSFT:
OK I see you didn't merge it yet, my mistake.  Let me make a pass. :)

lizap:
Thanks, Mike - merging now. You'll see the changes live this afternoon (12/20).

PR #2098, opened 2018-12-15
Correct Acronym SSH

gunnarhaslinger:


lizap:
Oh my, thanks for fixing this!

Issue #517, opened 2018-03-25
Recommended dimensions typo (350 should be 35 (I think))


JeremyTBradshaw:
This line:
"Important

We recommend the dimensions for the logo to be 260x350 @ 96 dpi with a file size of no greater than 10 KB."

I think it means to say 260x35.  I could be wrong.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 26b30a8b-731c-a11f-a666-83e25eef75ad
* Version Independent ID: 2f45486d-f1d0-bf45-3790-fedc30f5a658
* Content: [Changing the company logo on the AD FS sign-in page](https://docs.microsoft.com/en-us/windows-server/identity/ad-fs/operations/change-the-company-logo-on-the-ad-fs-sign-in-page)
* Content Source: [WindowsServerDocs/identity/ad-fs/operations/Change-the-company-logo-on-the-AD-FS-sign-in-page.md](https://github.com/MicrosoftDocs/windowsserverdocs/blob/master/WindowsServerDocs/identity/ad-fs/operations/Change-the-company-logo-on-the-AD-FS-sign-in-page.md)
* Service: **unspecified**
* GitHub Login: @billmath
* Microsoft Alias: **billmath**

Markus-Hanisch:
I agree. Annoying.
Correct value is stated here: Customizing the AD FS Sign-in Pages
https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2012-R2-and-2012/dn280950(v=ws.11)

doctordns:
As this has been addressed and a fix merged, can this issue be closed?

JeremyTBradshaw:
@doctordns Certainly!  Haha if the question is for me:).

Thanks very much!

illfated:
(ref. #929 (**Error in instructions?**) )

Issue #929, opened 2018-06-04
Error in instructions?


mdedmon:
I found this line in the style.css file in the default theme:

/* Logo image recommended dimension: 60x60 (square) or 350X35 (elongated), 4 kB average, 10 kB maximum. Transparent PNG strongly recommended. */

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 26b30a8b-731c-a11f-a666-83e25eef75ad
* Version Independent ID: 2f45486d-f1d0-bf45-3790-fedc30f5a658
* Content: [Changing the company logo on the AD FS sign-in page](https://docs.microsoft.com/en-us/windows-server/identity/ad-fs/operations/change-the-company-logo-on-the-ad-fs-sign-in-page)
* Content Source: [WindowsServerDocs/identity/ad-fs/operations/Change-the-company-logo-on-the-AD-FS-sign-in-page.md](https://github.com/MicrosoftDocs/windowsserverdocs/blob/master/WindowsServerDocs/identity/ad-fs/operations/Change-the-company-logo-on-the-AD-FS-sign-in-page.md)
* Product: **windows-server-threshold**
* GitHub Login: @billmath
* Microsoft Alias: **billmath**

illfated:
@mdedmon : I know, this comment is pretty late compared to when you posted this ticket, but do you recommend taking the text you quoted from the CSS and enter that in the "important" note instead of the current text, which is as follows:

> **We recommend the dimensions for the logo to be 260x35 @ 96 dpi with a file size of no greater than 10 KB.**

IngridAtMicrosoft:
#please-close

IngridAtMicrosoft:
#assign:IngridAtMicrosoft

PR #2109, opened 2018-12-18
Update Change-the-company-logo-on-the-AD-FS-sign-in-page.md

doctordns:
Changed note that logo should be 260x35, not 260x350

Resolves issue #517


illfated:
(ref. #929 (**Error in instructions?**) )

PR #2132, opened 2018-12-21
Change AIA example to HTTP as HTTPS is not supported

ph1ll:


lizap:
Thanks for catching this! We'll get the update published ASAP.

Issue #1935, opened 2018-11-15
Instructions to 'Open Hyper-V Manager' are confusing


olofh-msft:
The instructions say "Click Start, point to Administrative Tools, and then click Hyper-V Manager."
1) There is no 'Administrative Tools' to point to -- however there is a 'Windows Administrative Tools'
2) There is no 'Hyper-V Manager' under 'Windows Administrative Tools'

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: f6f91586-db15-ff49-b5e4-c22438f2ac76
* Version Independent ID: a9133113-eaba-89ee-477d-aa510ff9fe74
* Content: [The Hyper-V Virtual Machine Management service must be running](https://docs.microsoft.com/en-us/windows-server/virtualization/hyper-v/best-practices-analyzer/the-hyper-v-virtual-machine-management-service-must-be-running)
* Content Source: [WindowsServerDocs/virtualization/hyper-v/best-practices-analyzer/The-Hyper-V-Virtual-Machine-Management-service-must-be-running.md](https://github.com/MicrosoftDocs/windowsserverdocs/blob/master/WindowsServerDocs/virtualization/hyper-v/best-practices-analyzer/The-Hyper-V-Virtual-Machine-Management-service-must-be-running.md)
* Service: **na**
* Product: **windows-server-threshold**
* GitHub Login: @KBDAzure
* Microsoft Alias: **kathydav**

doctordns:
You need to first install the tools first - see the resolution section.

I have updated the instructions to clarify this.

doctordns:
To find the Hyper-V module, and to ensure you see it in the Start menu, you have to ensure the management tools are loaded. That means either performing the activity on the Hyper-V server or loading the RSAT Tools. 

See the resolution section on the page to get the management tools.

IngridAtMicrosoft:
#please-close

PR #2135, opened 2018-12-21
Update The-Hyper-V-Virtual-Machine-Management-service-must-be-running.md

doctordns:
Clarified that if you doin't have the tools installed, you won't see the GUI!

Addresses issue in #1935


Justinha:
@doctordns Thanks for the addition!

PR #2147, opened 2018-12-24
Fixed a misspelling

karammasri:
The word "own" was misspelled as "won"


lizap:
Thanks for catching this! We'll get the update published ASAP.

PR #2177, opened 2019-01-03
Update rds-supported-config.md

TomHickling:
Correcting the RDSH 2012 R2 column in the ## RD Connection Brokers table from Not Supported to Supported


lizap:
Thanks, Tom - looks like we had a merge conflict that wiped out your earlier edit. This will restore it.

PR #2303, opened 2019-01-26
Overhaul

skycommand:
This is a proposal of a rather dramatic change in the layout, which includes:

- Starting the page with an "Introduction" section that says what the Windows Admin Center is. This section was originally the "Windows Admin Center at a glance" section, which was buried near the bottom of the page.
- Following it up with a "Quick start" section. It is basically the "Get Started" section, which is moved and renamed. (It is renamed because one of its list items is called "getting started".)
- Following it up with the "Content at a glance" section. As before, it serves as the index for the whole Windows Admin Center library. However, the previous title of this section was "Project Honolulu is now Windows Admin Center!", which is a very strange title. As an index, it does not need to be front and center; the reader already finds another version of it docked to the left side of the screen.

  I've also adjusted the heading levels in this section to stop it from crowding the TOC to the right side of the page on Microsoft Docs.

In addition:

- The "Latest Features" section stays where it was, but its title is changed to "Release history", because judging by its contents, that's what it is. It does not confine or dedicate itself to the latest features.
- The last line of the article, which contains links to the Twitter account and the blog of Windows Admin Center is now optimized for accessibility. Previously, a screen reader for a blind person would read it:

  > "Follow us Follow us on Twitter Follow us on Blogs".

  Clearly, this is unfriendly. With my change, the screen reader reads:

  > "Stay updated! Follow us on Twitter. Visit our Blogs."

### DIFF
--- a/WindowsServerDocs/manage/windows-admin-center/overview.md
+++ b/WindowsServerDocs/manage/windows-admin-center/overview.md
@@ -15,17 +15,32 @@ ms.prod: windows-server-threshold
 
 Welcome to Windows Admin Center!
 
-**Windows Admin Center** is an evolution of Windows Server in-box management tools; it’s a single pane of glass that consolidates all aspects of local and remote server management. As a locally deployed, browser-based management experience, an Internet connection and Azure aren’t required. Windows Admin Center gives you full control of all aspects of your deployment, including private networks that aren’t Internet-connected.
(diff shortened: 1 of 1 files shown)


lizap:
@jwwool - can you take a look at these proposed changes to your WAC overview article? Thanks!

jwwool:
@lizap looks great! feel free to merge!

@skycommand Thanks for the contribution!

PR #2368, opened 2019-02-06
Chkdsk can be used only for local disks

outdat3dkernel:


outdat3dkernel:
It's obvious but it's nice to be listed :)

Justinha:
@outdat3dkernel thanks for the revision, and welcome to the list!

PR #2424, opened 2019-02-15
Update to Oracle Docs for 7.4 to 7.6 support

jackchammons:
Stole the columns from the RH docs
Updated the notes to match those on the oracle page


jackchammons:
@lizap

shirgall:
UEK5 supports SRIOV (Accelerated Networking) and has been released with Oracle Linux 7.6 and up. We should add a column for that.

jackchammons:
@lizap can we merge this?

Issue #785, opened 2018-05-10
AD FS Service Account requirements


jeremyhagan:
It isn't good enough to say "The permissions required at runtime will be added automatically when you configure AD FS." Some environments use group policy to override things like User Rights Assignments and so what ever is granted at runtime will be overwritten. Please detail the requirements in the documentation. 

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 68bac2aa-024c-a014-eeaf-f28a44b31c5e
* Version Independent ID: 86c83819-91d6-c990-16f1-d05ebdf07b58
* Content: [AD FS 2016 Requirements](https://docs.microsoft.com/en-us/windows-server/identity/ad-fs/overview/ad-fs-requirements)
* Content Source: [WindowsServerDocs/identity/ad-fs/overview/ad-fs-requirements.md](https://github.com/MicrosoftDocs/windowsserverdocs/blob/master/WindowsServerDocs/identity/ad-fs/overview/ad-fs-requirements.md)
* Service: **windows-server-threshold**
* GitHub Login: @billmath
* Microsoft Alias: **billmath**

illfated:
Ref. commit https://github.com/MicrosoftDocs/windowsserverdocs/commit/4e0bc9e0a99ae78c2e99d88d2588499686f71259 & PR https://github.com/MicrosoftDocs/windowsserverdocs/pull/2500

illfated:
#please-close

PR #2500, opened 2019-03-04
Add the current User Rights Assignment requirement

kenjohnson03:
These user rights assignments are required for the service to operate as expected.


illfated:
(ref. #785)

PR #2536, opened 2019-03-11
Win Srv 2019 DC min requirement 2008R2 -> 2008

RabanserD:
Changed Windows Server 2019 minimum functional level from 2008R2 to 2008, since it is possible to add a 2019 DC to a 2008 functional level (as long as DFS is used). No idea if it was a typo or on purpose 2008R2.


daveba:
@RabanserD Please resolve your merge conflict and resubmit so we can review.  Thanks!

RabanserD:
@daveba PR can be closed, since similar change has already been deployed.

PR #2541, opened 2019-03-11
Storage Replica is available in Std

allanhirt:
https://docs.microsoft.com/en-us/windows-server/storage/whats-new-in-storage

"Storage Replica in Windows Server, Standard Edition

You can now use Storage Replica with Windows Server, Standard Edition in addition to Datacenter Edition. Storage Replica running on Windows Server, Standard Edition, has the following limitations:

    Storage Replica replicates a single volume instead of an unlimited number of volumes.
    Volumes can have a size of up to 2 TB instead of an unlimited size."


daveba:
@JasonGerend, please take a look a this one when you get a chance.

JasonGerend:
Great catch, thanks Allan! @daveba, this is ready to merge.

PR #2548, opened 2019-03-12
Update Configuring-Alternate-Login-ID.md

alextok:
Having this registry key published result in abusing them, and more problems in future. We decided to kill them, put minimum requirement for Windows as RS3, and enforce clients to update on RS3+. For RS2 cases, we prefer solve on case by case bases.


billmath:
#sign-off

PR #2671, opened 2019-04-03
Fix typo in xcopy.md

CaseyCarter:
"excuded" => "excluded"


eross-msft:
Thanks for catching that!

PR #2750, opened 2019-04-23
Recommendation for adding Chrome to WIASupportedUserAgents is too broad

bspender:
The current recommendation seems to suggest adding the “Chrome” string to WIASupportedUserAgents, but I think that is too broadly defined and could lead to a match when it probably should not because all Chrome user agent strings start with “Mozilla/5.0” followed by the platform. Adding "Chrome" by itself would lead to scenarios where WIA could incorrectly be attempted for devices such as Chrome for Android, which has a user agent string such as:

Mozilla/5.0 (Linux; <Android Version>; <Build Tag etc.>) AppleWebKit/<WebKit Rev> (KHTML, like Gecko) Chrome/<Chrome Rev> Mobile Safari/<WebKit Rev>


daveba:
@billmath Please review and if you agree with the proposed change, please sign-off so I can merge.  Thanks!

billmath:
checking with engineering

bspender:
For what it's worth, I just fixed an issue where my customer encountered this because they had used "chrome" instead of the more specific variants. Their 3rd party [phone] app (which I believe was built on the Electron app framework, which commonly has "chrome" in it's user agent strings) didn't support a mixed mode environment and required forms based auth on devices like Android ...but it kept trying WIA and failing. Once we replaced "chrome" with the specific variants - it all worked like a champ... Feel free to ping me internally if you need more specific details.  *appreciated!

PR #2816, opened 2019-05-10
Move the "Deployment options" section to the top

skycommand:
This PR moves the "Deployment options" to the top because reading it is a prerequisite for reading the rest of the article. One must first know what is a desktop mode and what is a gateway mode before knowing which OS is supporting them.


skycommand:
So, I can't see the conflict report without repopening the PR? That's bad...

Issue #2186, opened 2019-01-04
Pre-requisites - ports needed 


jorlopama:
On the Pre-requisites section ,  it is stated that "Web app proxy is able to communticate(typo) with the following over ports 80 and 443" ,  this is misleading because the actual ADFS Web Application proxies (WAPS) do not need this requirement, the communication only goes from the "internal" ADFS servers (not the WAPs). That section should probably be removed since the note next to it also mentions network pre-requisites from ADFS Servers

At least it should be clarified that that these prerequisites are not for "ADFS Web app proxies"  but Egress Traffic Proxies (or Internet traffic filters) 

Also , all calls to Azure MFA are over port 443 only - no calls are going through port 80 so that needs to be corrected too

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: ae96cfcf-0a50-7e13-30c8-43e3730e184f
* Version Independent ID: 9d8ed7b7-eec6-1b95-40b8-24a13361f66d
* Content: [Configure AD FS 2016 and Azure MFA](https://docs.microsoft.com/en-us/windows-server/identity/ad-fs/operations/configure-ad-fs-and-azure-mfa)
* Content Source: [WindowsServerDocs/identity/ad-fs/operations/configure-ad-fs-and-azure-mfa.md](https://github.com/MicrosoftDocs/windowsserverdocs/blob/master/WindowsServerDocs/identity/ad-fs/operations/configure-ad-fs-and-azure-mfa.md)
* Product: **windows-server-threshold**
* GitHub Login: @billmath
* Microsoft Alias: **billmath**

jorlopama:
Documentation has been properly updated

PR #2832, opened 2019-05-14
Update to remove WAPs pre-requisite

jorlopama:
This is related to #2186  , confirmed with product team - ADFS Web application proxies pre-requisites is not needed and it was misleading - removing that section and also removed the port 80 pre-requisite from the note since PGs also confirmed all communication goes over https 443


billmath:
#sign-off

PRMerger6:
@daveba : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #2856, opened 2019-05-21
Generation 2 in Azure reference added

MikeBazMSFT:


daveba:
@MikeBazMSFT Please fix merge conflict and I will approve and publish this pull request.  Thanks!

MikeBazMSFT:
Hmm.  Generation 2 is still only preview as far as I know, which makes the use of "supported" in the other version potentially problematic.  But the two are basically saying the same thing so I'm not sure my change makes sense now, two months after proposing it, so I'm going to close and let the existing version stand.

PR #2887, opened 2019-05-29
License conversion additional info provided for Install | Upgrade | Migrate to Windows Server 2019

jeoviedo:
Added additional info at License Conversion to note that while doing a license conversion from Standard to Datacenter is supported and doable, Datacenter and Standard is not.

#sign-off


jeoviedo:
#sign-off

PR #2892, opened 2019-05-29
Added a better description of Kerberos by IP feature

SteveSyfuhs:
The original feature documentation wasn't particularly clear about how and why you would use this new capability. This doc is a starting point to clarify why you need it and how you use it.


SteveSyfuhs:
#sign-off

PR #2949, opened 2019-06-13
Create-a-Bootable-USB-Flash-Drive.md

yannara:
Diskpart's command "Active" does not work anymore in last Windows 10 builds. The error is always "The Selected disk is not fixed MBR disk. The Active command can only be used on dixed MBR disks". It doesn't matter, is it NTFS or FAT32. Without this, the USB disk will not boot. This worked before, but not anymore (I suspect something changed in Windows last year)


yannara:
Hello! I was wrong! Forget about my comments. Sorry about this.

PR #3186, opened 2019-08-06
Add SAC channels to supported guest OSes

cwilhit:
We support SAC channels of Windows Server as a guest operating system assuming they're within their 18-month lifetime window. I think this should get called out in the table (given customer comments on the article stating this is not clear)


cwilhit:
#sign-off

cwilhit:
Moving to private repo.

PR #3209, opened 2019-08-13
Update getting-started-with-group-managed-service-accounts.md

jeoviedo:
Updated to include a space on line 265 in between "ITFarm1" and "-PrinciPalsAllowedToRetrieveManagedPassword". If a person were to copy the code used as an example and replace names in there with their own, it would have failed due to the lack of a space between a parameter and a value.


jeoviedo:
#sign-off

PR #3228, opened 2019-08-19
fixed diskspd cmdline

nedpyle:
was missing a now-required flag (changed in later release of the tool)


JasonGerend:
This is ready to publish #sign-off

Issue #2485, opened 2019-02-28
sxstrace does not accept slashes


marc-groundctl:
Running the example commands does not work; sxstrace just prints out its help.  However, if the slashes are replaced with dashes, the commands work.  The documentation should be updated so the parameter syntax is correct.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 00064275-ab46-02df-3c98-adc4c2f08db4
* Version Independent ID: 18dee0ea-506d-83b7-02ca-cbf3589b64a2
* Content: [sxstrace](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/sxstrace)
* Content Source: [WindowsServerDocs/administration/windows-commands/sxstrace.md](https://github.com/MicrosoftDocs/windowsserverdocs/blob/master/WindowsServerDocs/administration/windows-commands/sxstrace.md)
* Product: **windows-server-threshold**
* GitHub Login: @coreyp-at-msft
* Microsoft Alias: **coreyp**

marc-groundctl:
Forgot to mention my Windows version: 10.0.17763.316

Science-and-Motion-Sports:
You are totally right, I have the same issue and solution.

IngridAtMicrosoft:
Thanks for your dedication to our documentation. If we have been unable to review your issue in a timely manner, we sincerely apologize for the delayed response. We have captured the informtion you have provided for review. We are closing this issue for now, but if you feel that it's still a concern, please respond and let us know. If you determine another possible update to our documentation, please don't hesitate to reach out again. #please-close

PR #3295, opened 2019-09-10
Update sxstrace.md

heathhenley:
The `/` style options were not accepted using cmd.exe shell in windows versions: `Win 10 Pro, 1809, OS build 7763.678` and `Win 7 pro version 6.1 build 7601 SP1`


heathhenley:
Related to #2485

PR #3376, opened 2019-10-07
Add space to registry key

PeterUpfold:
The `Windows NT` registry key was missing a space, and it was also referred to as a 'hive', instead of a 'key'.


illfated:
Good point, I am fairly sure that `HKEY_` is a shortened form of `HIVE_KEY` so that would mean that only the top level is a hive.
- https://docs.microsoft.com/en-us/windows/win32/sysinfo/registry-hives

PR #3419, opened 2019-10-18
clarification needed

rdiazel:
it took me a while to find out that this does not work in 2012R2. is not supported and CMDLETS are not available

### DIFF
--- a/WindowsServerDocs/identity/ad-fs/operations/Configure-AD-FS-to-authenticate-users-stored-in-LDAP-directories.md
+++ b/WindowsServerDocs/identity/ad-fs/operations/Configure-AD-FS-to-authenticate-users-stored-in-LDAP-directories.md
@@ -11,7 +11,7 @@ ms.prod: windows-server
 
 ms.technology: identity-adfs
 ---
-# Configure AD FS to authenticate users stored in LDAP directories in Windows Server 2016,2019.(2012R2 is not supported)
+# Configure AD FS to authenticate users stored in LDAP directories in Windows Server 2016 or later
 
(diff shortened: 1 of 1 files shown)


WindowsServerDocs/identity/ad-fs/operations/Configure-AD-FS-to-authenticate-users-stored-in-LDAP-directories.md
```diff
@@ -11,7 +11,7 @@ ms.prod: windows-server
 
 ms.technology: identity-adfs
 ---
-# Configure AD FS to authenticate users stored in LDAP directories
+# Configure AD FS to authenticate users stored in LDAP directories in Windows Server 2016,2019.(2012R2 is not supported)
```
illfated: just a simple suggestion of style & readability (spacing) improvement:
```suggestion
# Configure AD FS to authenticate users stored in LDAP directories in Windows Server 2016 and 2019 (2012 R2 is not supported)
```
Note that this is a header text (only 1 # in front), so a period character is not expected.

illfated:
Looking back at the whole document, it seems that this added text may fit better as an "Applies to ..." line beneath the title line.
I will wait for feedback from the document owner or manager for a confirmation about this.

eross-msft:
@billmath  Please review this text and let me know if it's ready to merge. Thanks!

illfated:
@rdiazel : I have updated my suggestion, in case you want to commit that change directly. You are of course free to choose which method you want to use to change the page title.

billmath:
@rdiazel Looks good, thanks!

#sign-off

PR #3420, opened 2019-10-19
Fix spelling mistake

ItzLevvie:
* Changed "form" to "from"


illfated:
Looks good to me, simple enough typo fix. Would it be OK to add a simple MarkDown spacing adjustment (for compatibility) too, since you are already editing this section?
```diff
- >If you currently use DirectAccess, [...]
+ > If you currently use DirectAccess, [...]
```
... and the same modification in the previous line:
```diff
- >[!TIP]
+ > [!TIP]
```
(if you find it problematic to re-edit your PR content, I can easily suggest the edits directly in the source text, then you can commit the changes here on this page.)

ItzLevvie:
cc @daveba

PR #3422, opened 2019-10-20
Update getting-started-with-group-managed-service-accounts.md

dacarroll:
Updated examples to using samaccountname versus simply the ADObject name string. Inputing a string in this field if not located in the default AD containers will fail without using the samaccountname, DN, or ObjectGUID.

### DIFF
--- a/WindowsServerDocs/security/group-managed-service-accounts/getting-started-with-group-managed-service-accounts.md
+++ b/WindowsServerDocs/security/group-managed-service-accounts/getting-started-with-group-managed-service-accounts.md
@@ -167,7 +167,7 @@ Membership in **Domain Admins**, **Account Operators** or ability to create msDS
     Enter the command on a single line, even though they might appear word-wrapped across several lines here because of formatting constraints.
 
     ```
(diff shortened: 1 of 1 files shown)


illfated:
These lines are PowerShell cmdlets, right? Would you mind improving the document by adding the syntax name `PowerShell` to the preceding line of back ticks for each cmdlet line you have edited?
For cosmetic purposes, it would also be nice if you could remove the surplus blank lines below many of the cmdlets.

(if you prefer, I can add the suggestions directly in this PR content, but I thought it would be better if you are able to make all the changes in 1 extra commit by yourself, instead of getting 1 commit for each change if I add suggestions for all the changes.)

dacarroll:
@illfated Good call. Updated.

illfated:
This topic is not my field of expertise, so I don't have any working experience with it. Even so, the http notation (`http/ITFarm1.contoso.com/contoso.com, http/ITFarm1.contoso.com/contoso, http/ITFarm1/contoso.com, http/ITFarm1/contoso`) looks a bit odd to me, but I presume there would have been complaints about incorrect command parameters from other users way before this, had they been incorrect in this format and usage.

PR #3424, opened 2019-10-21
Updating H3 to H2

kaushika-msft:
1. Figured the H3 headers were missing from the right-side navigation (In this article).
2. Title update


illfated:
Ref. PR #3486

PR #3426, opened 2019-10-21
Update Hyper-V and Debian versions

loozhengyuan:
**Updated Hyper-V Versions**
The documentation `Applies to` section is updated to include Server 2019 and Hyper-V 2019.

**Updated Debian Versions**
All Debian versions are updated to use `MAJOR.x` version naming convention to reduce the need to regularly update this section on the latest compatibility. In any case, this page is not regularly updated anyway, so keeping to this style will avoid giving readers the ambiguity as to whether a new minor version of Debian is supported by Hyper-V or not.


eross-msft:
This PR is causing conflicts which I can't fix. I'm closing this now. If there are still issues with this article, please open a new pull request. Thanks!

PR #3442, opened 2019-10-28
exit_2.md Capitalization update (exit command)

stefkiourk:


illfated:
Suggestion: add capitalization to the **Description** for `/b` under **Parameters**, if you are able to edit that part.
> exits the current batch script instead of exiting Cmd.exe. If executed from outside a batch script, exits Cmd.exe.

stefkiourk:
> Suggestion: add capitalization to the **Description** for `/b` under **Parameters**, if you are able to edit that part.
> 
> > exits the current batch script instead of exiting Cmd.exe. If executed from outside a batch script, exits Cmd.exe.

Done!

stefkiourk:
> ~Looks good to me. File content is limited, so I have nothing further to add.~
> 
> Scratch that. According to the public web page (https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/exit_2 ), the tagged parameter name `<exitCode>` is hidden because MarkDown does not display HTML tags within standard text.
> 
> @stefkiourk : Could you also add back ticks around the parameter `<exitCode>` in the small table?

Sure, done! Good observation. (Ironically I named the 

> ~Looks good to me. File content is limited, so I have nothing further to add.~
> 
> Scratch that. According to the public web page (https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/exit_2 ), the tagged parameter name `<exitCode>` is hidden because MarkDown does not display HTML tags within standard text.
> 
> @stefkiourk : Could you also add back ticks around the parameter `<exitCode>` in the small table?

Sure, done! Good observation. Messed the commit name a bit ("exitcode") but content is as you noted.

Issue #3542, opened 2019-11-22
Variable mistake in text


patrickvdw:
https://docs.microsoft.com/en-us/windows-server/identity/ad-fs/operations/configure-ad-fs-and-azure-mfa#renew-and-manage-ad-fs-azure-mfa-certificates.

In the section about renewing the MFA certificate the new certificate is using the variable: $newcert. However, in the text below is spoken of the a $certbase64 variable:
**command:**
_PS C:/> New-MsolServicePrincipalCredential -AppPrincipalId 981f26a1-7f43-403b-a875-f8b09b8cd720 -Type Asymmetric -Usage Verify -Value $newcert_
**text:**
_"$certbase64 is the new certificate. The base64 encoded certificate can be obtained by exporting the certificate"_

patrickvdw:
I've added a new pull request.

illfated:
PR #3544 **Update configure-ad-fs-and-azure-mfa.md** , resulting in commit https://github.com/MicrosoftDocs/windowsserverdocs/commit/c88c7f224d2fa42b20ac94478bc8dd4e012f27a5 .

illfated:
Issue resolved.

#please-close

PR #3544, opened 2019-11-22
Update configure-ad-fs-and-azure-mfa.md

patrickvdw:
Change $certbase64 to $newcert in section ### Configure each new AD FS Azure MFA certificate in the Azure AD tenant.


illfated:
(ref. #3542)

PR #3607, opened 2019-12-09
Adding supportability note

kaushika-msft:
After discussion with PGPM, SPMs and SMEs, adding supportability statement.


kaushika-msft:
#Sign-off

PR #3612, opened 2019-12-09
Fixed netsh command to change static IP Address

dpaulson45:
If you include "name" after "address" in the actual command you will get an error "The filename, directory name, or volume label syntax is incorrect."


illfated:
#sign-off

PR #3624, opened 2019-12-11
Adding WAC gifs to Azure Monitor docs

adagashe:


JasonGerend:
@daveba This looks ready to publish. 

Adi - the gifs might not render properly and it's hard to tell on the public site, so it'll be important to double-check once published.

adagashe:
Sounds good, I can double check it when it is live!

Thanks,
Adi

PR #3648, opened 2019-12-19
wrong spelling of an url

loveunicom:
### Check if Certificate Trusted issuers is configured properly in all the AD FS/WAP servers
*Common Symptom: HTTP 204 “No content from https\://certuath.adfs.contoso.com”*--->### Check if Certificate Trusted issuers is configured properly in all the AD FS/WAP servers
*Common Symptom: HTTP 204 “No content from https\://certauth.adfs.contoso.com”*


illfated:
#sign-off

Issue #486, opened 2018-03-16
Event Viewer Path Incorrect


jakemm:
Path in the document 'Applications and Services Logs\Microsoft\Windows\Microsoft\Authentication'
Actual path 'Applications and Services Logs\Microsoft\Windows\Authentication'

illfated:
Using the GitHub search feature, I can confirm that are no documents belonging to this repository containing the reported incorrect path, only the correct path.

I believe this issue has been resolved via a different ticket.

illfated:
#please-close

illfated:
#please-close

illfated:
#please-close

PR #3659, opened 2019-12-20
Fix Protected Users event logs location

cnotin:
Here is an example of the path:
image


illfated:
Looks good to me.

Protected Users Event Logs location

illfated:
Rich diff: https://github.com/MicrosoftDocs/windowsserverdocs/pull/3659/files?short_path=ef57df9#diff-ef57df968e91aa5f2db05e2061e69d8e

illfated:
(ref. #486)

PR #3669, opened 2019-12-21
Fix key file names, paths, typos...

alexandair:
Key file names are id_rsa and id_rsa.pub.
Lots of path were broken because of the way how `\.` was treated (removing `\`).

### DIFF
--- a/WindowsServerDocs/administration/OpenSSH/OpenSSH_KeyManagement.md
+++ b/WindowsServerDocs/administration/OpenSSH/OpenSSH_KeyManagement.md
@@ -12,7 +12,7 @@ ms.product: w10
 
 Most authentication in Windows environments is done with a username-password pair.
 This works well for systems that share a common domain. 
-When working across domains, such as between on-premise and cloud-hosted systems, it becomes more difficult.
+When working across domains, such as between on-premises and cloud-hosted systems, it becomes more difficult.
 
 By comparison, Linux environments commonly use public-key/private-key pairs to drive authentication.
 OpenSSH includes tools to help support this, specifically:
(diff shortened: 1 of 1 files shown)


onovotny:
What is the reasoning for changing from `ed25519` to `rsa`?

alexandair:
@onovotny That's what ssh-keygen generates by default: `id_rsa` and `id_rsa.pub`.

WindowsServerDocs/administration/OpenSSH/OpenSSH_KeyManagement.md
```diff
@@ -141,22 +141,22 @@ After completing these steps, whenever a private key is needed for authenticatio
 
 ## Deploying the public key
 
-To use the user key that was created above, the public key needs to be placed on the server into a text file called *authorized_keys* under users\username\.ssh\. 
+To use the user key that was created above, the public key needs to be placed on the server into a text file called *authorized_keys* under `C:\Users\username\.ssh\`. 
 The OpenSSH tools include scp, which is a secure file-transfer utility, to help with this.
 
-To move the contents of your public key (~\.ssh\id_ed25519.pub) into a text file called authorized_keys in ~\.ssh\ on your server/host.
+To move the contents of your public key (`~\.ssh\id_rsa.pub`) into a text file called authorized_keys in `~\.ssh\` on your server/host.
```
illfated: Maybe I don't understand how the procedure works, but it feels like this sentence is unfinished by not saying "what needs to be done" after the current content.

alexandair: I've only changed the name of the file and added inline backticks to format paths as a code and avoid publishing procedure to mess up with "\." and remove "\".

illfated: Very well, fair enough. I just thought maybe it could be useful to find out if the sentence was supposed to describe what kind of action is needed "to move the contents of your public key into a text file called authorized_keys on your server/host", because that logic seems to be missing. (I could very well be mistaken, I don't know this topic.)

alexandair: You are correct that that sentence doesn't sound like it's finished. Because it actually isn't. :(
It should be attached to the previous sentence. However, this whole section need to be rewritten because steps that are described are incorrect and won't allow a reader to authenticate using key files.

illfated: Thank you for the detailed reply. Fair enough. Maybe we could ask @maertendMSFT for a suggestion on how to deal with this particular question.

PR #3699, opened 2020-01-06
Updating landing page

kaushika-msft:
Added SMB links and made readable changes


illfated:
Wow, that is nice indeed. Relative links instead of direct external links (exiting to docs.microsoft.com when clicked while viewing the GitHub page).

I presume you have checked that all links are OK, but I will appreciate to have it confirmed.

PR #3700, opened 2020-01-06
Update mstsc.md 'v' parameter with missing bracket

jacklin213:
The /v flag was missing a closing > bracket in the Parameters section. 
Added bracket in the same location as shown in the Syntax section

### DIFF
--- a/WindowsServerDocs/administration/windows-commands/mstsc.md
+++ b/WindowsServerDocs/administration/windows-commands/mstsc.md
@@ -35,7 +35,7 @@ mstsc.exe /migrate
 |        Parameter        |                                                         Description                                                         |
 |-------------------------|-----------------------------------------------------------------------------------------------------------------------------|
 |    <Connection File>    |                                   Specifies the name of an .rdp file for the connection.                                    |
-|   /v:<Server[:<Port>]   |                Specifies the remote computer and, optionally, the port number to which you want to connect.                 |
+|   /v:<Server>[:<Port>]   |                Specifies the remote computer and, optionally, the port number to which you want to connect.                 |
 |         /admin          |                                   Connects you to a session for administering the server.                                   |
 |           /f            |                                    starts Remote Desktop Connection in full-screen mode.                                    |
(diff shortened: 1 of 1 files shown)


WindowsServerDocs/administration/windows-commands/mstsc.md
```diff
@@ -35,7 +35,7 @@ mstsc.exe /migrate
 |        Parameter        |                                                         Description                                                         |
 |-------------------------|-----------------------------------------------------------------------------------------------------------------------------|
 |    <Connection File>    |                                   Specifies the name of an .rdp file for the connection.                                    |
-|   /v:<Server[:<Port>]   |                Specifies the remote computer and, optionally, the port number to which you want to connect.                 |
+|   /v:<Server>[:<Port>]   |                Specifies the remote computer and, optionally, the port number to which you want to connect.                 |
```
illfated: if you look at the Rich diff view (https://github.com/MicrosoftDocs/windowsserverdocs/pull/3700/files?short_path=5525f76#diff-5525f76b28b2922a671f33a8b3272d60), it indicates that adding the &gt; bracket, it will hide the content within the brackets, as well as the brackets, treating the bracketed content as if it was some kind of HTML tag. I think these brackets need the addition of an escape character (backslash) to be kept visible in the document:
```suggestion
|   /v:<Server\>[:<Port\>]   |                Specifies the remote computer and, optionally, the port number to which you want to connect.                 |
```
(this effect can also be seen when viewing your result file https://github.com/MicrosoftDocs/windowsserverdocs/blob/127aab0a5a04d77aaa078d1193429a2af0cff602/WindowsServerDocs/administration/windows-commands/mstsc.md)

jacklin213: Great catch, apologies I didn't notice that earlier. Fixed in latest commit.
Just realized the ##Syntax section had it in a code block which is probably why it didn't need escape characters where as this one does

illfated: Very good. Feel free to mark this conversation as Resolved to minimize it.

PR #3711, opened 2020-01-08
Fixed date

EliotSeattle:


illfated:
I disagree with this change. The American (EN-US) date format is used without preceding zeros in the date notation.

eross-msft:
I'm going to accept and merge this change. Per the contributor's guide for GitHub, the ms.date metadata attribute should be formatted as MM/DD/YYYY.

PR #3738, opened 2020-01-15
Fix file path to cloud print deploy log file

louistlhong:
Fix file path to cloud print deploy log file

### DIFF
--- a/WindowsServerDocs/administration/hybrid-cloud-print/hybrid-cloud-print-deploy.md
+++ b/WindowsServerDocs/administration/hybrid-cloud-print/hybrid-cloud-print-deploy.md
@@ -211,7 +211,7 @@ To enable authenticated communication with the HCP services, we need to create 3
     ![Print Server Cloud Print Deploy](../media/hybrid-cloud-print/PrintServer-CloudPrintDeploy.png)
 
     - Check the log file to see if there is any error:
-    `C:\Program Files\WindowsPowerShell\Modules\PublishCloudPrinter\1.0.0.0>notepad CloudPrintDeploy.log`
+    `C:\Program Files\WindowsPowerShell\Modules\PublishCloudPrinter\1.0.0.0\CloudPrintDeploy.log`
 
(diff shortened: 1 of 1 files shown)


WindowsServerDocs/administration/hybrid-cloud-print/hybrid-cloud-print-deploy.md
```diff
@@ -211,7 +211,7 @@ To enable authenticated communication with the HCP services, we need to create 3
     ![Print Server Cloud Print Deploy](../media/hybrid-cloud-print/PrintServer-CloudPrintDeploy.png)
 
     - Check the log file to see if there is any error:
-    `C:\Program Files\WindowsPowerShell\Modules\PublishCloudPrinter\1.0.0.0>notepad CloudPrintDeploy.log`
+    `C:\Program Files\WindowsPowerShell\Modules\PublishCloudPrinter\1.0.0.0\CloudPrintDeploy.log`
```
illfated: Please note that the original text intends to show the reader how to open the log file in Notepad from within the directory folder
```
C:\Program Files\WindowsPowerShell\Modules\PublishCloudPrinter\1.0.0.0>
```
and to open the file `CloudPrintDeploy.log` there. Your suggested change will only work for those users who have configured Windows to open any `*.log` file in Notepad or a similar editor for viewing.

I do not recommend making this change.

illfated:
As an example, I can list a `*.log` file in its directory and open it by using only its filename, but that is because I have defined `.log` files to be a Notepad++ document:

image

Conversely, if I try to open some random `*.ETL` file, it will produce the following result:

image

image

illfated:
Of course, it could be said that the command would need to be surrounded with double quotes to be useful in a Windows Run command:

**`"C:\Program Files\WindowsPowerShell\Modules\PublishCloudPrinter\1.0.0.0>notepad CloudPrintDeploy.log"`**

Then again, if the reader is not aware of this requirement, the user would be quite a novice and should not be dealing with Azure Active Directory or deploying cloud print for servers. That is a job for a real IT administrator who knows what they are doing.

PR #3759, opened 2020-01-18
correction -She -replaced with you

NagaCSC:
I spot every where , it mentioned "She" where it suppose to be "you"


illfated:
No. That is incorrect. You have misunderstood the concept of 
## Scenarios

A scenario is an example, demonstration or description of what it could be like, with given circumstances. These particular scenarios are not related to the reader and should therefore not point to the reader when it comes to mentioning recommended improvements. These 5 lines are about a hypothetical company and their IT department. The pronoun "She" is being used to highlight that it does not have to be a male person doing the job.

Anyway, if you want to avoid being gender specific, it is possible to use "**they**" to be gender neutral.

eross-msft:
@NagaCSC  To be inclusive, Microsoft is moving away from "he/she" to "they/them". If you would like to go through and change to that, I can merge your changes.

NagaCSC:
I made those correction as suggested. Please advice

eross-msft:
@NagaCSC  I will go through the changes later today and if everything looks right, I'll go ahead and merge your changes. Thank you!!

PR #3842, opened 2020-02-05
Add details of using RRT to move from SQL to WID

maweeras-msft:
RRT can be useful for customers that want to simplify a SQL based farm configuration by moving to WID. Similarly a customer that grows an existing WID based farm to sizes requiring SQL, could do so using the RRT tool. RRT is often suggested via support cases to customers but we dont have this documented in public for customers to self discover.


eross-msft:
@billmath Please confirm if this text is okay to merge. Thanks!!

billmath:
#sign-off

PR #3902, opened 2020-02-15
Add note for 49443 port

juntakata:
One of my Premier customer requested to add a note regarding 49443 port. The customer could not successfully authenticate client because 49443 port was used by other services on WAP.


WindowsServerDocs/identity/ad-fs/design/AD-FS-Requirements.md
```diff
@@ -191,7 +191,10 @@ Configuring the following network services appropriately is critical for success
 Both the firewall located between the Web Application Proxy and the federation server farm and the firewall between the clients and the Web Application Proxy must have TCP port 443 enabled inbound.  
   
 In addition, if client user certificate authentication \(clientTLS authentication using X509 user certificates\) is required, AD FS in Windows Server 2012 R2 requires that TCP port 49443 be enabled inbound on the firewall between the clients and the Web Application Proxy. This is not required on the firewall between the Web Application Proxy and the federation servers\).  
-  
+
+> [!NOTE]
+> Please also make sure 49443 port is not used by other services on Web Application Proxy server.
```
illfated: This is a more fluent English way of saying/writing this sentence.
```suggestion
> Also make sure that port 49443 is not used by any other services on the Web Application Proxy server.
```
Removing "Please" here because the advice is not optional and may cause failure if not followed.

Issue #2570, opened 2019-03-15
Public key in TPM?  Looks wrong.


jordanmills:
"If Credential Guard is not available and a TPM is, then a public key is created protected by the TPM."

I think that should say that the private key is protected by the TPM.  The public key is intended to be shared freely, and protecting it is counterproductive.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 116f2435-8627-2501-f544-a80daee6b45d
* Version Independent ID: 85bb9ed3-f106-5cdc-f8c9-4efa018929a6
* Content: [Domain-joined Device Public Key Authentication](https://docs.microsoft.com/en-us/windows-server/security/kerberos/domain-joined-device-public-key-authentication)
* Content Source: [WindowsServerDocs/security/kerberos/domain-joined-device-public-key-authentication.md](https://github.com/MicrosoftDocs/windowsserverdocs/blob/master/WindowsServerDocs/security/kerberos/domain-joined-device-public-key-authentication.md)
* Product: **windows-server-threshold**
* GitHub Login: @michikos

illfated:
As you can see from the links directly above my comment, there was an attempt to clarify the text almost a year after this ticket, but it still needs to be more precise.

illfated:
PR #5154 is now ready for review, comments and suggestions.

IngridAtMicrosoft:
Thanks for your dedication to our documentation. If we have been unable to review your issue in a timely manner, we sincerely apologize for the delayed response. We have captured the informtion you have provided for review. We are closing this issue for now, but if you feel that it's still a concern, please respond and let us know. If you determine another possible update to our documentation, please don't hesitate to reach out again. #please-close

PR #3922, opened 2020-02-19
Update to clarify key generation

andrewwillett:
Clarifies a public/private key pair is generated, of which the private key is protected by Credential Guard or TPM.


illfated:
The text is still unclear, it still reads as if both keys are protected — ref. #2570 (**Public key in TPM? Looks wrong.**).

Never mind, I will open a new PR to clarify more precisely.

PR #4033, opened 2020-03-17
Update Supported Debian versions

linuxelf001:
9.0-9.6 to be replaced with 9.0-9.12
10 will be replaced with 10.0-10.3


eross-msft:
@linuxelf001  As we spoke about, please close this item and open in the private repo, windowsserverdocs-pr. I will approve and merge after it's there. Thanks!

PR #4035, opened 2020-03-18
SpecialPollInterval - Should this be build 1709?

peterbax117:
This is listed as the build 1702 which is not a build of Windows.  Is this just a typo and should be build 1709?
Line change is in 311


eross-msft:
@dcuomo  Please look at this change and let me know if it's ready to merge. Thanks!

illfated:
Just out of curiosity: even though 1709 is more widely known, maybe because of the title "Fall Creators Update" and because of the amount of changes, there is a slim chance that 1702 is a typo for 1703. Could you check which of those main versions (1703 or 1709) is the source of the first appearance of **SpecialPollInterval** in Windows 10?

peterbax117:
@illfated - Are you asking me this question? I have looked around publicly but we don't seem to list this addition as a feature anywhere.  The only place I have found this information is in this article.  May need to have someone check the bug or feature request to determine when it was added.  i do not have access to that information.

illfated:
@peterbax117 : Thank you for your detailed reply. I did intend to ask you, but I now realize that this question needs to be looked into further, either by the document author, the document/program manager or a knowledgeable tech writer.

Looking at the comment above mine now, it appears that the Sr. Program Manager should be well qualified to verify and comment on the factual information.

peterbax117:
@illfated = Thank you for the clarification :)  Okay, so waiting on the Sr. Program Manager to respond then.

illfated:
Yep, I am also looking forward to see if @dcuomo can share the necessary facts with us.

peterbax117:
@dcuomo - Any update here?

illfated:
I am starting to wonder if he needs a tweet on twitter to be reminded to visit GitHub once in a while.

dcuomo:
@eross-msft - Please don't contact me through this mechanism. If you need work related assistance, please send information to my internal email address so I can properly address it and work with my team to identify the correct answer if need be. I mute notifications in GitHub specifically because I get tons of them constantly and important ones like this get lost in the noise.

eross-msft:
@dcuomo  No problem now that I know. I believe there are others in here. I will look and send you an email with the associated issues if I find them.

dcuomo:
I believe this should be RS2 which did ship, so 1703 would be the correct number.

illfated:
Thanks a lot for the feedback, also for confirming my suspicion.

Issue #4047, opened 2020-03-20
DNS Suffixes


archgriffin:
Is listed to comma separate them, but it wants them semi-colon separated in the VPN profile XML file. 

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 77e14da1-102d-3ddf-e2a1-a20ce533b95a
* Version Independent ID: 374edc4c-f527-b7b7-c291-353928b97d8d
* Content: [Configure Windows 10 Client Always On VPN Connections](https://docs.microsoft.com/en-us/windows-server/remote/remote-access/vpn/always-on-vpn/deploy/vpn-deploy-client-vpn-connections#feedback)
* Content Source: [WindowsServerDocs/remote/remote-access/vpn/always-on-vpn/deploy/vpn-deploy-client-vpn-connections.md](https://github.com/MicrosoftDocs/windowsserverdocs/blob/master/WindowsServerDocs/remote/remote-access/vpn/always-on-vpn/deploy/vpn-deploy-client-vpn-connections.md)
* Product: **windows-server**
* Technology: **networking-ras**
* GitHub Login: @shortpatti
* Microsoft Alias: **pashort**

shortpatti:
Can a Microsoft employee take ownership of this topic?  I haven't worked there since Nov 2018.

hosebei:
On my system I was required to use comma (,). I have tested with a semi-colon (;) and the DNS Suffix entry is incorrectly listed. So it might be a Language/Cultural issue?
I have used en-US OS with de-CH settings (without LP)

shortpatti:
Can a Microsoft employee PLEASE remove my name from all the documents I used to own and manage? I signed the divorce papers from Microsoft in Nov 2018 and would LOVE IT if a Microsoft writer would take ownership of the content. I don't work there, and don't get paid to correct these documents anymore. I feel sorry for the IT Professionals using these products because there's no one at Microsoft who really cares about the IT Pro :( It's obvious because these documents haven't been maintained for over 18 months.

PLEASE PLEASE PLEASE someone remove my name from these documents!!!!! I can unsubscribe until my face turns blue, but that's 100s if not 1000s of documents.

illfated:
[...]

PR #4052, opened 2020-03-22
Updated Profile Name in PowerSHell Example

hosebei:
As outlined in the text above of the modification, if the "$Profilename" "has a space or other non-alphanumeric character, it must be properly escaped according to the URL encoding standard." But in the given example, there are just whitespaces. This commit does change that.


illfated:
By the way, do you have any opinion about the issue mentioned in ticket #4047 (**DNS Suffixes**) on the same page?

hosebei:
Regarding to #4047, I have tested both, comma and semi-colon, and did not receive an error. But I did only test to create the XML and load into the Windows 10 Client. I did not test if the dns suffixes properly work.

Issue #4051, opened 2020-03-22
Can the "Applies to" be updated to include Windows Server 2019


doctordns:
This page also applies to Server 2019.

Also, should Windows 8 removed?

---
#### Document details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 7dc50ab2-5b70-2b78-5345-cb261071fcc0
* Version Independent ID: c081e16d-cf9d-f9a7-e86c-80a38e33d211
* Content: [Hyper-V Virtual Machine Connection](https://docs.microsoft.com/en-gb/windows-server/virtualization/hyper-v/learn-more/hyper-v-virtual-machine-connect?redirectedfrom=MSDN#feedback)
* Content Source: [WindowsServerDocs/virtualization/hyper-v/learn-more/hyper-v-virtual-machine-connect.md](https://github.com/MicrosoftDocs/windowsserverdocs/blob/master/WindowsServerDocs/virtualization/hyper-v/learn-more/hyper-v-virtual-machine-connect.md)
* Product: **windows-server**
* Technology: **compute-hyper-v**
* GitHub Login: @KBDAzure
* Microsoft Alias: **kathydav**

illfated:
As long as the page documentation still applies to Windows 8.1 and Windows 8, I don't think it should be removed before Windows 8.1 and Windows 8 are no longer supported. There might still be valid Enterprise use of those Windows versions out there.

illfated:
@doctordns : Unless there is anything else you would like to change to complete a resolution of this issue, I think we can safely say that this ticket is resolved.

illfated:
#sign-off

doctordns:
So far as I can see, this is not resolved, in that Server 2019 has not been added?

illfated:
Fair enough. Opening a new PR to resolve that issue.

doctordns:
Should Server 2022 be added?

illfated:
> Should Server 2022 be added?

Way too early, I don't think it will be added before the version is publicly released to everyone.

doctordns:
The reason I ask is that Insiders Builds now have the Server 2022 branding as 20H1.

illfated:
I would not count on information being added before the release (RTM) version has been published, unless you are referring to Insiders Preview content pages.

PR #4053, opened 2020-03-22
Update hyper-v-virtual-machine-connect.md

doctordns:
Fixed typo with /Edit option.


illfated:
By the way, as you stated in issue ticket #4051 (**Can the "Applies to" be updated to include Windows Server 2019**), "This page also appies to Server 2019.", do you plan to incorporate that change in this PR, or do you prefer to make a separate PR for that change?

doctordns:
> By the way, as you stated in issue ticket #4051 (**Can the "Applies to" be updated to include Windows Server 2019**), "This page also applies to Server 2019.", do you plan to incorporate that change in this PR, or do you prefer to make a separate PR for that change?

I will leave those as separate issues if that is OK. I am not clear on what the policy is for changing the applies to tags.

Issue #4049, opened 2020-03-21
Please update page title, this is not a insider preview anymore 


carlosmayol:
Please update page title, this is not a insider preview anymore 

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 7ac2c66a-9734-5680-04e0-1aed2df2eb0e
* Version Independent ID: 3406e752-cef6-d640-db14-9836278f8ec3
* Content: [Insider Preview for HPN features in Windows Server 2019](https://docs.microsoft.com/en-us/windows-server/networking/technologies/hpn/hpn-insider-preview)
* Content Source: [WindowsServerDocs/networking/technologies/hpn/hpn-insider-preview.md](https://github.com/MicrosoftDocs/windowsserverdocs/blob/master/WindowsServerDocs/networking/technologies/hpn/hpn-insider-preview.md)
* Product: **windows-server**
* Technology: **networking**
* GitHub Login: @shortpatti
* Microsoft Alias: **pashort**

illfated:
Fair enough. Windows Server 2019 should be well out of its "Insider Preview" phase.
Do you have a suggestion for the page title?

I tried to synthesize a title based on the 2 sections in the page, but I am drawing the short straw here. I presume that is because I am not familiar enough with the content topics.

- [**Dynamic vRSS and VMMQ**](https://docs.microsoft.com/en-us/windows-server/networking/technologies/hpn/hpn-insider-preview#dynamic-vrss-and-vmmq)
- [**Receive Segment Coalescing (RSC) in the vSwitch**](https://docs.microsoft.com/en-us/windows-server/networking/technologies/hpn/hpn-insider-preview#receive-segment-coalescing-rsc-in-the-vswitch)

carlosmayol:
Yes:  New HPN Features in WS2019.

Thanks

Carlos Mayol Berral  | Sr. PFE | Infrastructure, Virtualization, Azure

carlosmayol:
New HPN Features in WS2019 should made it.

shortpatti:
Can a Microsoft employee please remove my name from these topics?  I signed the divorce papers in Nov 2018.  If this topic is not updated or supported anymore, then MICROSOFT should show some courtesy and tell their customers.

illfated:
[...]

PR #4060, opened 2020-03-23
Updating title for this page

carlosmayol:
as is not longer a insider preview


illfated:
Ref. issue ticket #4049

Issue #3404, opened 2019-10-15
Kernel parameter has been translated (elevator => Lift)


CeciAc:
Hi team, 
I moved this from here: https://github.com/MicrosoftDocs/windowsserverdocs.de-de/issues/5

The user reports this segment was over translated *elevator=noop*
The issue replicates in all locales. To fix this, please use the no-lock feature in this segment to avoid it to be machine translated

@DamianFekete commented 10 hours ago
A kernel parameter (passed as parameter in grub) was translated from elevator = noop to lift = noop.

He should not be translated.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 17bc364f-8fd0-b273-50bc-63a0d240f8d0
* Version Independent ID: dd4c402b-1f9f-f3d9-bb98-4b9c2c20f04a
* Content: [Best Practices for running Linux on Hyper-V](https://docs.microsoft.com/en-us/windows-server/virtualization/hyper-v/Best-Practices-for-running-Linux-on-Hyper-V#feedback)
* Content Source: [WindowsServerDocs/virtualization/hyper-v/Best-Practices-for-running-Linux-on-Hyper-V.md](https://github.com/MicrosoftDocs/windowsserverdocs/blob/master/WindowsServerDocs/virtualization/hyper-v/Best-Practices-for-running-Linux-on-Hyper-V.md)
* Service: **na**
* Product: **windows-server**
* Technology: **compute-hyper-v**
* GitHub Login: @shirgall
* Microsoft Alias: **kathydav**

illfated:
I would like to offer a PR to add MarkDown code block back ticks around the text, to avoid having that text translated. Adding MD back ticks around text has often helped when used on documents in the [MicrosoftDocs/windows-itpro-docs](https://github.com/MicrosoftDocs/windows-itpro-docs) repository where I have made some contributions, both in pull requests and answering comments in issue tickets.

illfated:
Resolved in commit https://github.com/MicrosoftDocs/windowsserverdocs/commit/82337afeb3798911903676fdd69295c53b44deb4 (new command structure).

illfated:
#please-close

PR #4212, opened 2020-04-22
From private repo

huypub:


illfated:
> (ref. #3404 Kernel parameter has been translated (elevator => Lift) )

PR #4248, opened 2020-04-28
Repair extensive damage to the list of links

skycommand:
Hello! 😊

I believe viewing my proposed changes in GitHub's Rich Edit view gives you the best idea as to what I did, but here is a summary anyway.

The list of links in this article was:
- Unformatted
- Stitched together without so much as a carriage return
- Corrupted, as some characters were replaced with a dash
- Impossible to properly read and use

I fixed all of those. The list should not appear as:
- Properly formatted as a bulleted list
- Each item consists of a single entry
- Slash characters restored
- Easy to browse and use


illfated:
Well done. Nice job.

PR #4261, opened 2020-04-30
Syntax get/set-adserviceaccount seems wrong

bartvermeersch:
See https://docs.microsoft.com/en-us/powershell/module/activedirectory/set-adserviceaccount?view=winserver2012-ps
and https://docs.microsoft.com/en-us/powershell/module/activedirectory/get-adserviceaccount?view=winserver2012-ps


illfated:
#sign-off

PR #4266, opened 2020-05-01
Update dcb-manage.md

vcozier:
Document is inconsistent with https://docs.microsoft.com/en-us/powershell/module/netqos/new-netqospolicy?view=win10-ps
For Create QoS Policy example 1, we are only naming the policy but not linking it to anything.  The example text talks about using a template.


illfated:
#sign-off

PR #3952, opened 2020-02-26
sc and sc.exe on powershell

OzBob:
updated examples

Issue #3822, opened 2020-01-30
PowerShell cannot run the sc delete command


tcHylmrX:
Tried the command in PowerShell in Administrator mode and the service is not removed and no error message returns.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 3ef8ad4c-710f-4a51-a0ed-cb2245f2051f
* Version Independent ID: b35a40fc-7798-c4ff-226c-20aad279fd39
* Content: [Sc delete](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/sc-delete#feedback)
* Content Source: [WindowsServerDocs/administration/windows-commands/sc-delete.md](https://github.com/MicrosoftDocs/windowsserverdocs/blob/master/WindowsServerDocs/administration/windows-commands/sc-delete.md)
* Product: **windows-server**
* Technology: **manage-windows-commands**
* GitHub Login: @coreyp-at-msft
* Microsoft Alias: **coreyp**

illfated:
Caused by the issue described in ticket #2898 (**Using `sc` in PowerShell has only unintended consequences**).

This ticket page could be a duplicate to the issue ticket linked above.

thallam08:
Use "sc.exe" to get Service Control instead of alias

illfated:
> Use "sc.exe" to get Service Control instead of alias

Quite so, but also see PR #3952 (**sc and sc.exe on powershell**) to resolve this and issue ticket #2898 (**Using `sc` in PowerShell has only unintended consequences**).

skycommand:
Resolved via #4290

illfated:
#sign-off

illfated:
#please-close

illfated:
#please-close

PR #4290, opened 2020-05-05
Improve sc.exe documentations

skycommand:
This pull request changes four pages:

- sc-config.md
- sc-create.md
- sc-delete.md
- sc-query.md

Changes include:

- Change all invocations containing `sc` to `sc.exe` because, in PowerShell, the former is an alias for `Set-Content`. This change has been discussed in #3952. This PR is the result of that discussion.
- Fill in the missing "Description" field with meaningful sentences. Instead of "Reference topic for \*\*\*\* - " it now says, e.g. "Learn how to change service configurations using the sc.exe utility."
- Change an incorrect remark about using the "Add or Remove Programs" module of Control Panel. Since Windows Server 2003 (released 17 years ago), this module is no longer in charge of configuring server roles. This change applies to `sc-delete.md` only.

### DIFF
--- a/WindowsServerDocs/administration/windows-commands/sc-config.md
+++ b/WindowsServerDocs/administration/windows-commands/sc-config.md
@@ -1,5 +1,5 @@
 ---
-title: Sc config
+title: Sc.exe config
 description: Reference topic for **** - 
 
 ms.prod: windows-server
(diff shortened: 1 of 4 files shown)


skycommand:
This PR potentially addresses #3822 and #2898 too.

illfated:
#sign-off

PR #4293, opened 2020-05-05
Ambiguous configuration

misejoh:
As it states that a system crash is likely to occur, IMO the right action is to put page file in a redundant disk.

### DIFF
--- a/WindowsServerDocs/administration/performance-tuning/hardware/index.md
+++ b/WindowsServerDocs/administration/performance-tuning/hardware/index.md
@@ -43,7 +43,7 @@ When your computer runs low on memory and it needs more immediately, Windows use
 You can optimize paging by using the following guidelines for page file placement:
 - Isolate the page file on its own storage device, or at least make sure it doesn't share the same storage devices as other frequently accessed files. For example, place the page file and operating system files on separate physical disk drives.
 
-- Place the page file on a drive that is fault-tolerant. If the disk fails, a system crash is likely to occur. If you place the page file on a fault-tolerant drive, remember that fault-tolerant systems are often slower to write data because they write data to multiple locations.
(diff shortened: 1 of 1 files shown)


illfated:
I could be mistaken, but I thought fault-tolerant drives are meant to mitigate system crashes if a single disk fails.

Current original text:
> Place the page file on a drive that is not fault-tolerant. If the disk fails, a system crash is likely to occur.

To me, this makes more sense than to remove "not" from the first sentence. Then again, there could easily be something I am missing.

PRMerger13:
@misejoh : Thanks for your contribution! The author(s) have been notified to review your proposed change. @phstee

WindowsServerDocs/administration/performance-tuning/hardware/index.md
```diff
@@ -43,7 +43,7 @@ When your computer runs low on memory and it needs more immediately, Windows use
 You can optimize paging by using the following guidelines for page file placement:
 - Isolate the page file on its own storage device, or at least make sure it doesn't share the same storage devices as other frequently accessed files. For example, place the page file and operating system files on separate physical disk drives.
 
-- Place the page file on a drive that is not fault-tolerant. If the disk fails, a system crash is likely to occur. If you place the page file on a fault-tolerant drive, remember that fault-tolerant systems are often slower to write data because they write data to multiple locations.
+- Place the page file on a drive that is fault-tolerant. If the disk fails, a system crash is likely to occur. If you place the page file on a fault-tolerant drive, remember that fault-tolerant systems are often slower to write data because they write data to multiple locations.
```
illfated, line 46: Based on your suggested change, I propose the following additional change:
```suggestion
- Place the page file on a drive that is fault-tolerant. If a non-fault-tolerant disk fails, a system crash is likely to occur. If you place the page file on a fault-tolerant drive, remember that fault-tolerant systems are often slower to write data because they write data to multiple locations.
```

PRMerger16:
@eross-msft : Thanks for your contribution! The author(s) have been notified to review your proposed change. @phstee

PR #4314, opened 2020-05-09
Correct the %* variable name

phuclv90:
It's `%*`, not `%\*`


illfated:
BTW, this is the second last (or 4th out of 5) bullet point in this section:
- https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/shift#remarks

illfated:
#sign-off

PR #4329, opened 2020-05-12
TLSExtensions xmlns Incorrectly Contains "HTTPS"

amorrowbellarmine:
While attempting to test SSO with always-on VPN I kept running into issues with the profile failing to be installed through both Intune and PowerShell. I compared a working manual configuration against the example code showed on this page and the only difference was the URI in the TLSExtensions. They should be "http://..." instead of "https://..."

The examples from https://docs.microsoft.com/en-us/windows/client-management/mdm/vpnv2-csp also show "http" instead of "https".


illfated:
Good point. It is likely that commit https://github.com/MicrosoftDocs/windowsserverdocs/commit/ef412f6c9f8e7f2802a3274bebb7ac1105c124b1 went a little too far by including the URLs in this page.

illfated:
#sign-off

PR #4361, opened 2020-05-17
About Limit number of connections policy

tayasuta:
When RDS client session is blocked by the Limit number of connections policy, the following message will appear : 
"The number of connections to this computer is limited and all connections are in use right now. Try connecting later or contact your system administrator."
Users do not receive "Remote Desktop Service is currently busy" message as described in this docs. So I think we should remove this section.

### DIFF
--- a/WindowsServerDocs/remote/remote-desktop-services/troubleshoot/remote-desktop-service-currently-busy.md
+++ b/WindowsServerDocs/remote/remote-desktop-services/troubleshoot/remote-desktop-service-currently-busy.md
@@ -47,24 +47,5 @@ To resolve this issue, apply the following updates to the RDSH servers:
   - Windows Server 2012 R2: KB 4343891, [August 30, 2018—KB4343891 (Preview of Monthly Rollup)](https://support.microsoft.com/help/4343891/windows-81-update-kb4343891)
   - Windows Server 2016: KB 4343884, [August 30, 2018—KB4343884 (OS Build 14393.2457)](https://support.microsoft.com/help/4343884/windows-10-update-kb4343884)
 
-## Check the connection limit policy
 
-You can set the limit on the number of simultaneous remote desktop connections at the individual computer level or by configuring a group policy object (GPO). By default, the limit is not set.
-
-To check the current settings and identify any existing GPOs on the RDSH server, open a command prompt window as an administrator and enter the following command:
-  
-```cmd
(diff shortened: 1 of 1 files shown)


illfated:
Are you sure about this selection? The selection you are suggesting to remove does not match the description you have written. Or have I misunderstood your description completely?

You have selected to remove the last section, containing 2 commands (which look useful to me, not related to the message "Remote Desktop Service is currently busy").

The section you are suggesting to remove looks like this:
***
## Check the connection limit policy

You can set the limit on the number of simultaneous remote desktop connections at the individual computer level or by configuring a group policy object (GPO). By default, the limit is not set.

To check the current settings and identify any existing GPOs on the RDSH server, open a command prompt window as an administrator and enter the following command:

```cmd
gpresult /H c:\gpresult.html
```

After this command finishes, open **gpresult.html**. In **Computer Configuration\\Administrative Templates\\Windows Components\\Remote Desktop Services\\Remote Desktop Session Host\\Connections**, find the **Limit number of connections** policy.

  - If the setting for this policy is **Disabled**, then group policy is not limiting RDP connections.
  - If the setting for this policy is **Enabled**, then check **Winning GPO**. If you need to remove or change the connection limit, edit this GPO.

To enforce policy changes, open a command prompt window on the affected computer, and enter the following command:

```cmd
gpupdate /force

tayasuta:
@Trond B. Krokli<mailto:notifications@github.com>

Yes. As far as I know, when  the number of simultaneous remote desktop connections are limited due to this policy, the message "Remote Desktop Service is currently busy" does not appear. So I don't think this section about the connection limit policy exists on this page is appropriate.
I suggest we remove this section and create new page about the message "The number of connections to this computer is limited and all connections are in use right now. Try connecting later or contact your system administrator." for this policy.

- tatsuro

illfated:
@eross-msft - for your information, I have nothing further to add here.

eross-msft:
@tayasuta @illfated  Thank you for your contributions! I will check with the product group and get back with you about this.

eross-msft:
@tayasuta @illfated Thanks for your patience. I spoke with the product group and apparently, this message still appears for older, but still supported, versions of Windows Server. Because of that I'm going to have to close your change request. With that being said, knowing that newer versions don't have the same messaging, I've put an item in the writing backlog about creating a new article to address this issue for newer operating system versions. Thanks again!

PR #4432, opened 2020-06-01
ADFSDocUpdate

ryanpodonnell1:
Original article stated to place an ADFS server in front of a firewall (i.e. not protected by one). This was a tad bit confusing and should be updated to be placed `"behind"` a firewall


billmath:
#sign-off

PR #4437, opened 2020-06-02
Update dir.md - define /r

adzm:
/r is an option visible in /? help for displaying alternate data streams.

### DIFF
--- a/WindowsServerDocs/administration/windows-commands/dir.md
+++ b/WindowsServerDocs/administration/windows-commands/dir.md
@@ -20,7 +20,7 @@ The **dir** command can also run from the Windows Recovery Console, using differ
 ## Syntax
 
 ```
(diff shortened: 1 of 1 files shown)


illfated:
By the way, I have been pondering whether these Windows Commands pages should be updated to modern Windows Server / Windows 10 character format, where most, if not all of the help texts, show the parameters as uppercase characters, as well as in a slightly different order. On one hand, it would be nice to see the documentation following the real life view of the commands, but on the other hand, it is quite an amount of pages to go through to make this change.

eross-msft:
@illfated  Here are my thoughts about the upper/lowercase issue. So long as the information is accurate and people can use the commands/parameters, there are much larger issues to deal with. For example,  there is quite a bit of incorrect or completely missing text.
As I'm going through the Windows Commands, I'm actually choosing to make the commands/parameters mostly lowercase, as that's typically what we've done in the past in documentation, but really it's just a choice. 
There are 100's of command-related files. In this case, I'd prefer your help with the more fundamental issues rather than the grammatical issues....

PR #4492, opened 2020-06-11
Add missing word

rbyshko:


WindowsServerDocs/administration/OpenSSH/OpenSSH_KeyManagement.md
```diff
@@ -27,7 +27,7 @@ If you are unfamiliar with SSH key management, we strongly recommend you review
...
 
-SSH public-key authentication uses asymmetric cryptographic algorithms to generate two key files – one "private" and the other "public". The private key files are the equivalent of a password, and should protected under all circumstances. If someone acquires your private key, they can log in as you to any SSH server you have access to. The public key is what is placed on the SSH server, and may be shared without compromising the private key.
+SSH public-key authentication uses asymmetric cryptographic algorithms to generate two key files – one "private" and the other "public". The private key files are the equivalent of a password, and should stay protected under all circumstances. If someone acquires your private key, they can log in as you to any SSH server you have access to. The public key is what is placed on the SSH server, and may be shared without compromising the private key.
 
 When using key authentication with an SSH server, the SSH server and client compare the public key for username  provided against the private key. If the public key cannot be validated against the client-side private key, authentication fails.
```
illfated, line 32: Suggesting an improvement to this line as well, not anything missing as such, but it would be nice to be specific about these details:
```suggestion
When using key authentication with an SSH server, the SSH server and client compare the public keys for username provided against the private key. If the server-side public key cannot be validated against the client-side private key, authentication fails.
```
(Adding the plural form of "key" because they are the keys on the client and the server, even though they ought to be the same. Removing 2 redundant blanks as well.)

PR #4493, opened 2020-06-11
Fix mistake due to markdown rendering

rbyshko:


WindowsServerDocs/administration/OpenSSH/OpenSSH_KeyManagement.md
```diff
@@ -140,7 +140,7 @@ After completing these steps, whenever a private key is needed for authenticatio
 
 ## Deploying the public key
 
-To use the user key that was created above, the public key needs to be placed on the server into a text file called *authorized_keys* under users\username\.ssh\. 
+To use the user key that was created above, the public key needs to be placed on the server into a text file called *authorized_keys* under users\username\\.ssh\.
```
illfated, line 143: Good start, but the problem moves 1 word ahead, so the next backslash needs to be escaped too:
```suggestion
To use the user key that was created above, the public key needs to be placed on the server into a text file called *authorized_keys* under users\username\\.ssh\\.
```
(removing a blank space at end-of-line too)

rbyshko:
Right, thanks for improvement.

illfated:
If you also would be so kind to _merge_ the suggested change, it will complete the PR (as I see it).

rbyshko:
@illfated Sure, I have merged your changes.

PR #4537, opened 2020-06-21
removes duplicate content

thabaum:
Fixes #4573

duplicate information removed

image

Is this correct format?  I am not sure what `|LinksXML|` with the `|` is suppose to do to the document, it looks confusing I wanted to fix it just didn't know if it was by design.


eross-msft:
@thabaum  This file is more messed up than most. The pipe is used to show tables, but it looks like somehow this table was broken. I'm going to merge this PR since the duplicated content needs to go, but I'm going to add a backlog item to look at this file more closely and to fix it up. Thank you for catching this.

thabaum:
Makes sense, I was trying to figure out what was happening on this section of the page and how I could make it look right.  I figured you would catch it and understand better than I would what was to be the expected result.  Thanks for the tip on the pipe for showing tables.  Learn something new everyday :)

eross-msft:
Please continue to send pull requests....there are LOTS of fun things we continue to learn about markdown :D

PR #4588, opened 2020-06-24
Enable Windows 10 Home Remote Desktop

marcusnestor:
Remote Desktop allows you to connect and control your device from a remote device with the use of Remote Desktop client. You can work from another device as if you were working directly on your computer.


eross-msft:
The product team has stated that this pull request must be declined, as it: 1) Includes content from a third-party site and 2) isn't an approved topic as it directs users to install a third-party tool that enables RD usage in Win10 Home edition, which is an unsupported feature.

PR #4591, opened 2020-06-25
Fixed DNS instruction issues

garrett-wood:
instructions indicated a CNAME record should be created but specified an IP address. CNAME records for the adding an additional name a DNS name goes by, but A records are intended for IP addresses.

Also fixed minor grammar / syntax issues, and added a note specifying that if IPv6 was used, an additional AAAA record should be created.


billmath:
#sign-off

PR #4594, opened 2020-06-26
Update deploy-roaming-user-profiles.md

jg2703:


eross-msft:
@jg2703 Thank you for taking the time to suggest this change; however, we do the release versions from newest to oldest. Therefore, it needs to stay Windows Server 2019 and then Windows Server 2016. Thanks!

PR #4613, opened 2020-07-04
ConfigurationVersion related info

githubarpyka:
Many administrators try to upgrade the ConfigurationVersion to Windows Server 2019, however such ConfigurationVersion does not exist. WAP 2016 / 2019 use Windows Server 2016 as ConfigurationVersion.  Double cheked in my test environment and in the source code. We need to avoid more confusion here and I suggest that you accept my update request.

### DIFF
--- a/WindowsServerDocs/identity/ad-fs/deployment/upgrading-to-ad-fs-in-windows-server.md
+++ b/WindowsServerDocs/identity/ad-fs/deployment/upgrading-to-ad-fs-in-windows-server.md
@@ -139,6 +139,9 @@ Check the WAP configuration by running the Get-WebApplicationProxyConfiguration 
 ```PowerShell
 Get-WebApplicationProxyConfiguration
 ```
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


WindowsServerDocs/identity/ad-fs/deployment/upgrading-to-ad-fs-in-windows-server.md
```diff
@@ -139,6 +139,9 @@ Check the WAP configuration by running the Get-WebApplicationProxyConfiguration
 ```PowerShell
 Get-WebApplicationProxyConfiguration
 ```
+> [!NOTE]
```
illfated, line 142: You may want to add a blank line between previous formatting code and the new one:
```suggestion

> [!NOTE]
```
This is mostly a safety & codestyle advice. Even though the current content may work well, it is mostly because of supported compatibility.

illfated:
Otherwise, the content seems quite OK to me. I don't know enough about the details, but it looks straight-forward correct to me.

billmath:
#sign-off

PR #4625, opened 2020-07-09
mount.md: Escape backslash used as example

Erisa:
Previously used `(**\**)` which shows up as `(****)` on the web. Changing the `\` to `\\` resolves this.


WindowsServerDocs/administration/windows-commands/mount.md
```diff
@@ -38,7 +38,7 @@ mount [-o <option>[...]] [-u:<username>] [-p:{<password> | *}] {\\<computername>
...
 | -o lang=`{euc-jp|euc-tw|euc-kr|shift-jis|Big5|Ksc5601|Gb2312-80|Ansi)` | Specifies the language encoding to configure on an NFS share. You can use only one language on the share. This value can include any of the following values:<ul><li>**euc-jp:** Japanese</li><li>**euc-tw:** Chinese</li><li>**euc-kr:** Korean</li><li>**shift-jis:** Japanese</li><li>**Big5:** Chinese</li><li>**Ksc5601:** Korean</li><li>**Gb2312-80:** Simplified Chinese</li><li>**Ansi:** ANSI-encoded</li></ul> |
-| -u:`<username>` | Specifies the user name to use for mounting the share. If *username* isn't preceded by a backslash (**\**), it's treated as a UNIX user name. |
+| -u:`<username>` | Specifies the user name to use for mounting the share. If *username* isn't preceded by a backslash (**\\**), it's treated as a UNIX user name. |
```
illfated, line 41: Looks good so far. Since we are already editing this line, I think we should add spaces before & after the backslash to make it more visible ( within the parentheses ) :
```suggestion
| -u:`<username>` | Specifies the user name to use for mounting the share. If *username* isn't preceded by a backslash ( **\\** ), it's treated as a UNIX user name. |
```

Erisa: Sure thing. On the same page there's also instances where it shows an example of an asterisk, maybe it would be good to add spaces there too?

Erisa: Though perhaps out of the scope of this PR.

illfated: Fair point, although to me it seems like less of a problem with the asterisks, because they seem to have some minimal whitespace around them anyway. The backslash seems more problematic, because it joins the parentheses like a ligature, making it look like (\\) is a single symbol or character.

illfated:
#sign-off

WindowsServerDocs/administration/windows-commands/mount.md
Erisa: Ah I see, that's alright then, thanks!

illfated:
- [X] Rich diff view: https://github.com/MicrosoftDocs/windowsserverdocs/pull/4625/files?short_path=0d983d4#diff-0d983d4e1a3605d59beeae60b08beeff (easier to observe the effective changes)
- [X] Regular view: https://github.com/MicrosoftDocs/windowsserverdocs/pull/4625/files

PR #4675, opened 2020-07-28
Corrected unnecessarily escaped character

garrett-wood:
Unnecessarily escaped hyphen was causing backslash to appear in article title.


illfated:
#sign-off

PR #4688, opened 2020-08-01
Update create-failover-cluster.md

injadhav:
Certificate Backup Considerations after deploying AD Detached Cluster

### DIFF
--- a/WindowsServerDocs/failover-clustering/create-failover-cluster.md
+++ b/WindowsServerDocs/failover-clustering/create-failover-cluster.md
@@ -214,6 +214,8 @@ New-Cluster -Name CN=MyCluster,OU=Cluster,DC=Contoso,DC=com -Node Server1, Serve
 
 For examples of how to add clustered roles, see topics such as [Add-ClusterFileServerRole](https://docs.microsoft.com/powershell/module/failoverclusters/add-clusterfileserverrole?view=win10-ps) and [Add-ClusterGenericApplicationRole](https://docs.microsoft.com/powershell/module/failoverclusters/add-clustergenericapplicationrole?view=win10-ps).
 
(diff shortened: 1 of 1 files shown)


WindowsServerDocs/failover-clustering/create-failover-cluster.md
```diff
@@ -214,6 +214,8 @@ New-Cluster -Name CN=MyCluster,OU=Cluster,DC=Contoso,DC=com -Node Server1, Serve
 
 For examples of how to add clustered roles, see topics such as [Add-ClusterFileServerRole](https://docs.microsoft.com/powershell/module/failoverclusters/add-clusterfileserverrole?view=win10-ps) and [Add-ClusterGenericApplicationRole](https://docs.microsoft.com/powershell/module/failoverclusters/add-clustergenericapplicationrole?view=win10-ps).
 
+After the AD Detached failover Cluster is created backup the certificate with private key exportable option. Open MMC ==>File ==>Add remove Snap in ==>Certificates==>Services Accounts==>Next==>Local Computer==>Cluster Service==>Certificates==>Clussvc\Personal==>Select Certificate right click==>export ==>Next==>Yes export the Private Key ==>PfX Format==>Choose Password or you can add group ==>Next==>Select path where you want to store certificate==>Next ==>Finish. "More details have been documented in Internal Article https://internal.support.services.microsoft.com/en-US/help/4569891"
```
illfated, line 217: I am fairly sure we should do our best to apply the codestyle already present in this document (unless the MS Docs reviewer says otherwise). Anyway, my suggestion is as follows:
```suggestion
After the AD Detached failover Cluster is created, backup the certificate with the private key exportable option: Open **MMC** and go to the **File** menu. Select **Add/Remove Snap-in**, go to **Certificates**, then **Services Accounts** and click **Next**. Select **Local Computer** > **Cluster Service** > **Certificates** > **Clussvc\Personal** > **Select Certificate**. Right click, then select **export**, click **Next**. Press **Yes, export the Private Key**, select **PfX Format**, then **Choose Password**, or you can go to **Add group**, then click **Next**, then select the path where you want to store the certificate. Click **Next** and click **Finish**. "More details have been documented in Internal Article https://internal.support.services.microsoft.com/en-US/help/4569891"
```
I must admit that I am not entirely happy with my current result. This is more like a pointer to the direction in which I think this paragraph should be moving.
Also, I don't know why the last sentence should be enclosed in quotation marks.

@JasonGerend : Do you have any recommendations here?

JohnMarlin-MSFT: I believe this is from this doc

https://docs.microsoft.com/en-us/windows-server/failover-clustering/create-failover-cluster

This information that is being proposed changing does not exist in the actual doc.

eross-msft:
Confirmed with product team, this is ready to publish.

PR #4692, opened 2020-08-03
Fix an incorrect code snippet in ad-fs-build-custom-auth-method.md

yufeih:


billmath:
@yufeih Thanks for the update.  I approved it but there is a confilct.  Can your resolve this.  Thanks!

eross-msft:
@billmath @yufeih  I was actually able to fix the conflict before I merged, so we should be good. Thank you both!

PR #4779, opened 2020-09-02
Cleaned up AD FS custom auth development documentation

themattman18:
Removed duplicated code and changed code blocks to csharp blocks
Removed random periods in gacutil commands

### DIFF
--- a/WindowsServerDocs/identity/ad-fs/development/ad-fs-build-custom-auth-method.md
+++ b/WindowsServerDocs/identity/ad-fs/development/ad-fs-build-custom-auth-method.md
@@ -198,52 +198,6 @@ class MyPresentationForm : IAdapterPresentationForm
  }
 ```
 
-Now you should be able to F12 (right click – Go To Definition) on IAuthenticationAdapter to see the set of required interface members.
-Next, you can do a simple implementation of these.
-
-9. Replace the entire contents of your class with the following:
-
-```csharp
-    namespace MFAadapter
-    {
-    class MyAdapter : IAuthenticationAdapter
-    {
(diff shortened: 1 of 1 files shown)


billmath:
Checking with engineering.

billmath:
Hi @themattman18 

Unfortunately I haven't heard back from engineering yet.  So, I am going to go ahead and create a work item to track this.
For now, I am going to go ahead and close the PR and I will re-open it once I have heard back.

Thanks!

Bill

#please-close

illfated:
The review process should be quite a lot easier if you use the view **Hide whitespace changes** (https://github.com/MicrosoftDocs/windowsserverdocs/pull/4779/files?diff=unified&w=1) to see that most of the changes are formatting improvements.

The formatting improvements are straightforward and can be approved quite easily.

(I tried the rich diff view, but it leaves me with more questions than answers vs. the duplication removal, because this topic is not within my area of expertise.)

The large section of code deduplication is a bit more cumbersome for me to review, since I don't do any coding worth mentioning.
This is about the 2 big red blocks (Lines 203-222 & Lines 224-344):  https://github.com/MicrosoftDocs/windowsserverdocs/pull/4779/files?diff=unified&w=1#diff-95188fd5c2ced5f1a2b6e90c61177c2ef06e0f8586128c37fe9b438e02d69474L203-L344

To be able to verify that it is OK to remove that part, I would have to know this topic well enough to use it straight away as a lab work example.

I can only assume that the suggestion to remove the unique (for the page) text in Lines 203-207 is based on the fact that the code following step 9 is already used in the preceding examples (this is the part which could be easily verified by someone familiar with this particular form of scripting).

PR #4834, opened 2020-09-21
Fix table layout for cmdkey doc

filhit:


eross-msft:
@filhit  Thank you for taking the time to make this update; however, the change was actually caught last week, and is part of a pull request that will be published this week. I'm going to close this as it would cause a merge conflict to merge it, but please continue to create pull requests going forward!

PR #4878, opened 2020-10-04
Update OpenSSH_KeyManagement.md

McSpooder:
Update the docs to be more accurate as to why rsa keys should be used.


illfated:
(ref. #4897)

Issue #4639, opened 2020-07-15
Dead link to static IP address section


KidElectric:
title says it all
[Enter feedback here]

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 4fb3496f-9df6-8859-4569-ea033daed5ac
* Version Independent ID: 9f762b67-575d-ea11-08d5-edbab9a21f15
* Content: [Remote Desktop - Allow access to your PC from outside your network](https://docs.microsoft.com/en-us/windows-server/remote/remote-desktop-services/clients/remote-desktop-allow-outside-access)
* Content Source: [WindowsServerDocs/remote/remote-desktop-services/clients/remote-desktop-allow-outside-access.md](https://github.com/MicrosoftDocs/windowsserverdocs/blob/master/WindowsServerDocs/remote/remote-desktop-services/clients/remote-desktop-allow-outside-access.md)
* Product: **windows-server**
* Technology: **remote-desktop-services**
* GitHub Login: @Haley-Rowland
* Microsoft Alias: **elizapo**

RAJU2529:
@KidElectric. issue is conformed by me also after checking the link.

illfated:
<details>

@RAJU2529 : I think you mean **confirmed** with the letter **i** , instead of "conformed" (which means to comply with, align with or adapt to).

illfated:
- Archived page from 2019-04-11:
https://web.archive.org/web/20190411124602/https://docs.microsoft.com/en-us/windows-hardware/customize/mobile/mcsf/enable-static-ip
- Archived page from 2019-08-15:
https://web.archive.org/web/20190815232553/https://docs.microsoft.com/en-us/windows-hardware/customize/mobile/mcsf/enable-static-ip

illfated:
The entire section named "Customizations for mobile devices" has been removed from the Windows Hardware Development pages. I wonder if this has got anything to do with Windows 10 Phone not being developed any more, or if the pages simply have been moved to some other subsection of Internet of Things (IoT).

illfated:
Then again, there is the multitude of Microsoft Azure pages which also contains the topic "static IP address", but those are usually about more advanced services, like setting up Azure Kubernetes Service (AKS) or Azure Kubernetes Service (AKS) Load Balancer.

exchange12rocks:
[...]

PR #5088, opened 2020-12-07
Corrects #4639

exchange12rocks:
The dead link has never been useful for this article - it's about configuring mobile devices by OEM manufactures - completely different process,
So we can safely remove it


PRMerger9:
@exchange12rocks : Thanks for your contribution! The author(s) have been notified to review your proposed change.

eross-msft (approved):

PR #5093, opened 2020-12-08
Update cipher.md

michha:
Added the **_suspected_** overwriting method.
From using the command it outputs first zeros, then ones and at last random numbers.
If someone from the product team could have a look at the code and confirm my assumption please.

If I should add the algorithm (0, 1, random) instead of mentioning the DoD name I could make an additional change.


PRMerger10:
@michha : Thanks for your contribution! The author(s) have been notified to review your proposed change.

illfated (approved):
Looks OK to me, but it needs review and confirmation from the product team, as suggested by the PR author.

MicrosoftGuyJFlo:
As published in the product that is not listed. If the PG wants to update the command then it will update in the product help file.

MicrosoftGuyJFlo:
#please-close

PR #5151, opened 2020-12-28
Windows Commands/extract: add info about Extrac32

illfated:
**Description:**
As noted in issue ticket MicrosoftDocs#1369 (**no está disponible en Windows 10**), the command `extract` is not available in Windows 10 (and confirmed to be not available in Windows Server 2016).

Thanks to @PabloJoaquin for pointing out and reporting the issue.

**Changes proposed:**

- add [Note] blob with information about `Extract.exe` being replaced in Server 2016 (and newer) and Windows 10
- add syntax and parameter table copied from the output of `Extrac32 /? | more` in the same pattern as for `extract`
- add extra [Note] blob describing how to  redirect (filter) the extrac32 help command output via [more]
- add link to the command page for [more], within the sentence on how to filter the output via [more]
- add H2 headings for Extract.exe and Extrac32.exe
- add **/ extrac32** in the page title

**Ticket closure or reference:**
Closes MicrosoftDocs#1369

### DIFF
--- a/WindowsServerDocs/administration/windows-commands/extract.md
+++ b/WindowsServerDocs/administration/windows-commands/extract.md
@@ -6,14 +6,20 @@ ms.assetid: 20dab03e-f6e1-4eb8-b8a1-fd6f1d97ee83
 ms.author: lizross
 author: eross-msft
 manager: mtillman
-ms.date: 10/16/2017
+ms.date: 12/29/2020
 ---
 
-# extract
+# extract / extrac32
 
 Extracts files from a cabinet or source.
 
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


PRMerger19:
@illfated : Thanks for your contribution! The author(s) have been notified to review your proposed change.

illfated (commented):
"Hide whitespace changes" does not show up any different than Normal view.
"Rich diff view" is scarcely useful, since the added first Note blob makes GitHub show the existing text about Exctract.exe as new text, while some of the new text is shown as old or only modified.

eross-msft (approved):

PR #5160, opened 2020-12-31
Fixes typo (left should be right).

MicahZoltu:
image

Screenshot included for reference, this option is on the right.


PRMerger13:
@MicahZoltu : Thanks for your contribution! The author(s) have been notified to review your proposed change.

illfated:
Well, that is true for modern versions of Windows 10, but there is a caveat at the beginning of the document:

https://github.com/MicrosoftDocs/windowsserverdocs/blob/master/WindowsServerDocs/remote/remote-desktop-services/rds-vdi-recommendations.md#L21

```
These instructions and recommended settings are relevant to Windows 10 1607 (version 10.0.1393).
```

If the description of where menus are located is changed, it might be necessary to change other descriptions as well, or at least update the Windows version in the line mentioned above. I defer to the Microsoft Docs team for verification and feedback on this.

illfated:
Just to follow up on my previous comment, I was going to suggest that maybe there should be a minimum version number based on the supported build versions. The problem then becomes to check what else has changed in the complete document content. I don't know what would be a practical solution here. Maybe there could be a Note blob referencing in which Windows 10 Version/OS Build the change from left to right happened. What do you think?

PR #5162, opened 2020-12-31
Changes to disable short name generation.

MicahZoltu:
I believe enabling short names *hurts* performance, which is mostly what this document is about.  I am guessing this was just a mistake and it was intended that short names be disabled.

From the group policy text:
> Some applications require short names for compatibility, but short names have a negative performance impact on the system.


PRMerger20:
@MicahZoltu : Thanks for your contribution! The author(s) have been notified to review your proposed change.

illfated (approved):
Looks valid to me, but I defer to the document author or the MS Docs team for final verification.

PR #5194, opened 2021-01-13
Update Single-Page-Application-with-AD-FS.md

shekeriev:
There was a misleading entry for clientId. It was something like 'https://localhost:44326/', instead of the actual Client Id - a string in the form of '150ab73e-0b05-4b78-9e50-0095a992cca9'


PRMerger7:
@shekeriev : Thanks for your contribution! The author(s) have been notified to review your proposed change.

illfated (approved):
Thank you for the update and correction. Nice. This also looks a lot cleaner, from a code viewing perspective. Good job.

PR #5238, opened 2021-01-26
Update forfiles.md

6Anasovich9:
Using \*.\* as a mask lists files and not directories, we should use * instead or omit the /M parameter completely (as * is the default mask).


PRMerger18:
@6Anasovich9 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

illfated (approved):
Looks good to me.

eross-msft (approved):

PR #5274, opened 2021-02-05
Update cmdkey.md - fix backtick placement in delete

mgrandi:
it looks sorta wonky with the backtick being there, so i moved it over. 

also fix the pipe not being escaped

Before: 

2021-02-04 17_15_50 - Screenshot - cmdkey docs - before

After:

2021-02-04 17_16_18 - Screenshot - cmdkey docs - after

Also, I'm unsure if this will impact the rendered build on docs.microsoft.com, but the rendered build on github was messing up because of the unescaped pipe character. Butting a backslash before it to escape it works in the github rendered view at least, but this can be removed if it breaks the view on docs.microsoft.com


PRMerger10:
@mgrandi : Thanks for your contribution! The author(s) have been notified to review your proposed change.

illfated:
Interesting solution. Looks like it will work even on docs.microsoft.com — to my surprise. I did not expect the backslash escape character to work the same way within the MD backtick encapsulation.

illfated (approved):
I approve. Looks good to me.

mgrandi:
@illfated yeah, i think markdown is still kinda fragmented with parsing rules (commonmark is the closest thing we have), but i googled around and found several examples that said that escaping certain characters works within a fenced code block / backticks,, because you also need to be able to escape the actual backtick character

i saw this from 2017 as an example: https://github.com/dotnet/csharplang/pull/743

PR #5351, opened 2021-03-08
Added Airlock Secure Access Hub

daniestermann:
Please add Airlock Secure Access Hub as a product supporting AD FS. Thanks!

### DIFF
--- a/WindowsServerDocs/identity/ad-fs/operations/Configure-Additional-Authentication-Methods-for-AD-FS.md
+++ b/WindowsServerDocs/identity/ad-fs/operations/Configure-Additional-Authentication-Methods-for-AD-FS.md
@@ -25,6 +25,7 @@ Below is an alphabetical list of Microsoft and third-party providers with MFA of
 
 |Provider|Offering|Link to learn more|
 |-|-|-|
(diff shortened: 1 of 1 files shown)


PRMerger20:
@daniestermann : Thanks for your contribution! The author(s) have been notified to review your proposed change.

illfated:
@billmath : I could be mistaken, but I would think that adding external URLs in Microsoft Docs documentation pages should go through the appropriate partner program channels.

Are you available to confirm either my assumption or this PR content?

daniestermann:
> @billmath : I could be mistaken, but I would think that adding external URLs in Microsoft Docs documentation pages should go through the appropriate partner program channels.
> Are you available to confirm either my assumption or this PR content?

I am product marketer at Airlock and we just wanted to list our product here. I am not sure which is the correct way to do this. 
Can you guide us @billmath?

sethmanheim:
@billmath Is it OK to add this?

sethmanheim:
Unable to receive approval for this, so unfortunately need to close. Please work through different channels if you still want this to be added. cc @billmath

PR #5362, opened 2021-03-11
Update find.md

npocmaka:
This is one of the commands with the most ...hmm not-totally-accurate documentation.

double quotes in the searched string are mandatory. I've added them in the syntax description as they are in the help message. Also added them in the examples.

Also I don't think examples with doc files are good idea as they are not text files.

it is not true that the command does not accept wildcards in file names so I removed the part with the FOR command.

I've added the exit codes - I think it is good idea to have them for all the commands (when possible - sometimes you can't give a good description for all of them)

the example with searching all the files in a drive will be too slow and I think it's better to use a directory instead.


PRMerger7:

npocmaka:
By the way - what files I need to change If want to add a documentation about a new commands?  A lot are missing but the most obvious are some internal CMD.exe commands.

illfated:
> Is there a policy that do NOT recommend external links to places like stackoverflow,ss64.com and so on?

Yes, the general rule is to only use microsoft.com links or relative links to neighboring pages. Theoretically, you could use links to verified Microsoft affiliates, but that is a marginal area and probably needs to be verified by the product group to be approved.

> is there a way to put a link to a Microsoft's page?

Not quite sure what you mean, but you can usually add links to relevant microsoft.com pages, as long as the link is important for the topic in question.

PRMerger12:

illfated (approved):
Rich Diff view, to review using the layout view:
- https://github.com/MicrosoftDocs/windowsserverdocs/pull/5362/files?diff=unified&w=1&short_path=d780cc9#diff-d780cc93c09323fd1d4912d3b696fa3e05edac88d495eb28e6007b319e57673f

Looks quite good to me.

illfated:
#sign-off

PRMerger13:
@illfated: I'm sorry - only the author of this article, @JasonGerend, can sign off on your changes. But we do have an exception process - if you are on the Microsoft content or product team for this product area, you can ask the PR review team to review and merge it by sending mail to the techdocprs alias.

illfated:
Yes, I know. My sign-off comment is based on a previous agreement with the team. @khdownie : Would you happen to know if there are someone else available instead of @JasonGerend, since it seems like he is more than occupied with other work than these Pull Request?

khdownie:
@illfated I have a meeting with Jason on Monday morning and I will bring up the fact that we have a lot of these PRs that are being submitted through the public repo to correct issues, and then they just sit there. I know there are many others. I can't sign off on this as I'm not an author, but there has to be a way to get these (and other) corrections made.

illfated:
Thank you very much. I am relieved to hear that you are aware of the situation and that you will discuss what can be done. Best of luck to you, take care and stay safe. 👍🏼

JasonGerend (approved):

JasonGerend:
Looks good - sorry for the delay - thank you for your contributions!

illfated:
Oh. I am just relieved to see you back. I was worried for a while. Thank you to all of you.

PR #5428, opened 2021-04-03
Update shift.md example

yarcat:
Before this change `cmd.exe` would print an error if no args were provided.

### DIFF
--- a/WindowsServerDocs/administration/windows-commands/shift.md
+++ b/WindowsServerDocs/administration/windows-commands/shift.md
@@ -51,7 +51,7 @@ rem mycopy dir file1 file2 ...
 set todir=%1
 :getfile
 shift
-if %1== goto end
+if "%1"=="" goto end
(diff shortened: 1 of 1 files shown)


PRMerger12:
@yarcat : Thanks for your contribution! The author(s) have been notified to review your proposed change.

illfated (approved):
Good correction. The common rule of thumb is that batch files should always use double quotes for % parameters to accommodate for spaces in the argument.

illfated:
#sign-off

PRMerger20:
@illfated: I'm sorry - only the author of this article, @JasonGerend, can sign off on your changes. But we do have an exception process - if you are on the Microsoft content or product team for this product area, you can ask the PR review team to review and merge it by sending mail to the techdocprs alias.

illfated:
Yes, I know. My sign-off comment is based on an agreement with the team.

PR #5563, opened 2021-05-24
Per domain setting

rdiazel:
Hi team, I have added just a little of information as per a customer request as this was not clear for them and so it may not be clear to many people


PRMerger15:
@rdiazel : Thanks for your contribution! The author(s) have been notified to review your proposed change.

illfated:
Adding text prefaced with "Important:" is pretty close to merit its own [!IMPORTANT] colored blob.
- When the text content actually is important, it should be given its own colored note blob.
- If it does not require its own note blob, rephrase the text and the word "important".

PRMerger7:
@sethmanheim : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #5637, opened 2021-06-21
Update high-availability.md

fazars:
For #### Install with a self-signed certificate you need add command "-static address local ip"


PRMerger6:
@fazars : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #5641, opened 2021-06-22
Update troubleshooting.md

ManfredHelber:
The description in the updated section was not correct - you dont have to add the sites in the Internet Options but in the pup-up Settings in Edge


PRMerger14:
@ManfredHelber : Thanks for your contribution! The author(s) have been notified to review your proposed change.

EricBergDE:
Have seen the same issue. I agree with Manfred that the Edge setting is solving the issue.

PR #5648, opened 2021-06-25
fix order of calls in enableextensions example

knilch0r:
previously, the "verify other" would set errorlevel to 1 in any case.


PRMerger18:
@knilch0r : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #5655, opened 2021-06-28
Clarified path to event log for auditing

cunninghamp:
Clarified the path to the event log that records the audit events. Previous wording was unclear as to precise location of the logs.


PRMerger9:
@cunninghamp : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #5693, opened 2021-07-08
Update cscript.md

npocmaka:
'Option' is the term used for the command line switches used for the windows script host engine (at least in the help message) ,but not a parameter. Simple examples added.


PRMerger16:

PR #5722, opened 2021-07-20
Fix command for sshd autostart

saschanaz:
The current one autostarts ssh-agent instead.


PRMerger7:
@saschanaz : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #5738, opened 2021-07-25
Indicate VMName as variable

amiosci:
`VMName` as a variable without valid variable syntax is misleading as well as inconsistent with remaining documentation.


PRMerger9:
@amiosci : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #5752, opened 2021-07-28

DaPadrePedro:
Create only works if the named value doesn't exist already. By default the named value isn't there, but if it is (because of earlier administrative configurations for example), then it won't be changed! For example: suppose some admin configuration has set the named value with value 1 and later you want to set it to 0 through GPP, then it won't be set to 0 (disabled), because of Create. Taking Update as the action has the desired result though: update it when it already exists and if it doesn't exist yet, create it.
PS: Create is not always a wrong choice (in theory it could be the case that you only want to change the behavior, when the named value doesn't exist yet, although this seems a very exceptional scenario), but typically Update will cover the desired result better.


Deland-Han:
#sign-off

PR #5756, opened 2021-07-29

DaPadrePedro:
The registry key does exist by default; it's the named value that doesn't exist by default.


PRMerger16:
@DaPadrePedro : Thanks for your contribution! The author(s) have been notified to review your proposed change.

IngridAtMicrosoft:
@DaPadrePedro will you please check your assumptions with @nedpyle? The feedback I got from engineering is that registry value is actually correct.

PR #5771, opened 2021-08-04
ssh: Clarify the comment mkdir'ing the global ssh state directory

lf-:
It seems like it was copy pasted from elsewhere in the article and was not quite right for the context.


PRMerger15:
@lf- : Thanks for your contribution! The author(s) have been notified to review your proposed change.

robinharwood:
#assign:robinharwood

Hi @lf- ,

Firstly, thank you for your contribution and dedication to our documentation. I absolutely agree with your comment, good catch. In order to keep the language consistent with the article would you be happy to amend the text in your PR to the following?

# Make sure that the ssh directory exists in your server's program data folder

Thanks,
Robin

PRMerger12:
Users robinharwood are already assigned.

robinharwood:
#label:awaiting-response

PR #5781, opened 2021-08-06
Remove Disabling NCSI Passive Polling

tayasuta:
Disabling NCSI passive polling can cause "No Internet Access" unexpectedly. I do not recommend that we disable this feature. If this help for VDI optimization, instead of disabling this, I suggest that we change the interval.


PRMerger8:
@tayasuta : Thanks for your contribution! The author(s) have been notified to review your proposed change.

takayamm:
I agree to tayasuta.
My customer disables NCSI Passive Polling according to this KB.
Then network connectivity has become NoTraffic and BITS is not available in their environment.
So we are pleased to announce to enable NCSI Passive Polling to them.
Rave Case #26454002

a-hiiwam:
@Heidilohr @PRMerger8 Is this correction acceptable? May I ask your opinion?

kaushika-msft (approved):
Disabling NCIS passive polling is not recommended.
CC: @IngridAtMicrosoft

IngridAtMicrosoft (approved):

PR #5895, opened 2021-09-21
Add description for acronym (VDL)

tksh164:
`VDL` is not a general acronym, I think. So, it is need original spelling.

### DIFF
--- a/WindowsServerDocs/storage/refs/refs-overview.md
+++ b/WindowsServerDocs/storage/refs/refs-overview.md
@@ -26,7 +26,7 @@ ReFS introduces new features that can precisely detect corruptions and also fix 
 
 ### Performance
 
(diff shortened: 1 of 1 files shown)


PRMerger13:
@tksh164 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

JohnCobb1:
Hi @tksh164 Good call to define this acronym on first mention. In this case, the format should be "valid data length (VDL)" per this Microsoft source: https://docs.microsoft.com/en-us/windows/win32/fileio/file-streams

Please update your PR with this change, and then we can get it published. Thanks

PRMerger14:
@tksh164 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

tksh164:
@JohnCobb1 Thank you for your comment. Done the update my PR. Thanks.

PR #5906, opened 2021-09-25
Correct PowerShell example for `sshd` service

Sophismata:
The example code has the `ssh-agent` service being set to start automatically, *not* `sshd`.


PRMerger8:
@Sophismata : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Issue #5922, opened 2021-10-01
netsh http add sslcert example doesn't work


bcage29:
https://docs.microsoft.com/en-us/windows-server/networking/technologies/netsh/netsh-http#add-sslcert

Following is an example of the add sslcert command.

**Incorrect**
add sslcert ipport=1.1.1.1:443 certhash=0102030405060708090A0B0C0D0E0F1011121314 appid={00112233-4455-6677-8899- AABBCCDDEEFF}

**Correct**
add sslcert ipport=1.1.1.1:443 certhash=0102030405060708090A0B0C0D0E0F1011121314 appid='{00112233-4455-6677-8899- AABBCCDDEEFF}'

The appid parameter needs single quotes around the GUID, otherwise it fails and says "The parameter is incorrect."

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 48a10fe2-2374-889f-cc3b-4b078f9a2418
* Version Independent ID: 9e6e1e81-ad12-ac11-9464-4c5c8f6b192c
* Content: [Netsh Commands for Hypertext Transfer Protocol (HTTP)](https://docs.microsoft.com/en-us/windows-server/networking/technologies/netsh/netsh-http)
* Content Source: [WindowsServerDocs/networking/technologies/netsh/netsh-http.md](https://github.com/MicrosoftDocs/windowsserverdocs/blob/master/WindowsServerDocs/networking/technologies/netsh/netsh-http.md)
* Product: **windows-server**
* Technology: **networking**
* GitHub Login: @JasonGerend
* Microsoft Alias: **jgerend**

robinharwood:
We hugely appreciate your contribution, feedback, and support to our technical documentation.

We’ve phased out GitHub Issues on this repository as the feedback mechanism for content and have replaced it with a new feedback system. For more information see, [Content User Feedback](https://aka.ms/ContentUserFeedback). As part of this rollout, we are migrating your issue where our team will use your comments to improve the documentation experience. Unfortunately, this does mean you won't be able to access this issue after the migration is complete.

All issues will be migrated by Thursday 29th August 2024.

Once again, thank you for your valuable feedback, we hope you’ll continue to contribute using our new feedback system.

PR #5923, opened 2021-10-01
Added single quotes to fix netsh http add sslcert example

bcage29:
The netsh http add sslcert example fails because the appid is not contained within single quotes. Updated the example to have single quotes.

#5922


PRMerger14:
@bcage29 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #5928, opened 2021-10-02
Fix glitch regarding /Q parameter

HelderMagalhaes:
From format.com help:
/Q              Performs a quick format. Note that this switch overrides /P.
This is also consistent with the /P parameter description.


PRMerger18:
@HelderMagalhaes : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Issue #5704, opened 2021-07-13
"OpenSSH key management" corrections and feedback


abstrask:
## Host key generation section
https://docs.microsoft.com/en-us/windows-server/administration/openssh/openssh_keymanagement#host-key-generation

The `sshd` and `ssh-agent` is mixed up in the code sample. Fixed code:

```powershell
# Set the sshd service to be started automatically
Get-Service -Name sshd | Set-Service -StartupType Automatic

# Now start the sshd service
Start-Service sshd
```

## User key generation
https://docs.microsoft.com/en-us/windows-server/administration/openssh/openssh_keymanagement#user-key-generation

In the code sample, the `ssh-agent` service is set to start manually. Why not "automatic", so we don't have to manually start the SSH agent?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 4c89b7b3-c05f-b517-6109-04e6a97195c3
* Version Independent ID: 1e4b742e-d999-2c1e-a396-dc149f7ac76b
* Content: [OpenSSH key management for Windows](https://docs.microsoft.com/en-us/windows-server/administration/openssh/openssh_keymanagement)
* Content Source: [WindowsServerDocs/administration/OpenSSH/OpenSSH_KeyManagement.md](https://github.com/MicrosoftDocs/windowsserverdocs/blob/master/WindowsServerDocs/administration/OpenSSH/OpenSSH_KeyManagement.md)
* Product: **windows-server**
* Technology: **administration**
* GitHub Login: @maertendMSFT
* Microsoft Alias: **damaerte**

abstrask:
The "Host key generation" have been fixed with dc3187d0bef1246eb9b5a67feb59a7e22b734c0b.

robinharwood:
Hi @abstrask,

Thanks for your dedication to our documentation. I can see your point point was already close. Regarding the second, We've submitted a documentation change to update the code example, setting the ssh-agent to automatic, the updated article should be live by tomorrow. 

Thanks,

#please-close

PR #5969, opened 2021-10-15
Configure `ssh-agent` to start automatically

abstrask:
The documentation states:

> After completing these steps, whenever a private key is needed for authentication from this client, ssh-agent will automatically retrieve the local private key and pass it to your SSH client.

However, for this to work with current instructions, you need to manually start the agent after every reboot. This PR suggests to configure `ssh-agent` to start automatically instead, as proposed in #5704.


PRMerger6:
@abstrask : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #5975, opened 2021-10-18
Completed a fragmented sentence

danielharkerdp:


dknappettmsft:
Thanks for the feedback. I've corrected this in a slightly different way and the change should be visible later today.

#please-close

PR #6012, opened 2021-10-31
Incorrect statement about using SMS code

luisfeliz79:
RDP Gateway + NPS Azure MFA can only do Phone Call or Phone notification.  The article suggests that SMS code is possible, which is not the case.


PRMerger8:
@luisfeliz79 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6034, opened 2021-11-08
Update smb-direct.md

v-Andre:
Proposed change to correct note regarding teaming support for RDMA. LBFO does not support but SET does.
This note was probably taken from 2012 article (before SET).


PRMerger14:
@v-Andre : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6049, opened 2021-11-11
Fix formatting of DPPEval Tool

prasadu-microsoft:
Fixes the formatting of the output so it is more readable and likely true to the actual output.


PRMerger18:
@prasadu-microsoft : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6052, opened 2021-11-11
Update file-server-smb-overview.md

DaPadrePedro:
1) Rephrasing Windows versions to be more precise.
2) Tip: why not change the article a little bit to include Windows client? In practice Windows client is mentioned everywhere, except in the title and 'Applies to'.
3) Sometimes Updated seems more correct than New.
4) I've added the new features for W11/WS22.
5) Some minor details.


PRMerger12:
@DaPadrePedro : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Issue #2016, opened 2018-12-02
/E ?


Yetoo1:
What features does /E extend. /E is listed in more /? but not here.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: ccd04f7d-4ee2-0b51-fb21-30f9f0f21280
* Version Independent ID: c95145ee-6623-89f7-2254-1398f0a849aa
* Content: [more](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/more)
* Content Source: [WindowsServerDocs/administration/windows-commands/more.md](https://github.com/MicrosoftDocs/windowsserverdocs/blob/master/WindowsServerDocs/administration/windows-commands/more.md)
* Product: **windows-server-threshold**
* GitHub Login: @coreyp-at-msft
* Microsoft Alias: **coreyp**

vouser:
I have a response that may give you a useful answer. Skip to the end for it. But first I will pontificate.


Microsoft's documentation of Windows Commands is inexcusably bad.


1) There is no documentation of Windows Commands for Windows 
    10. Or Windows 8 that I can find. Producing documentation of 
    Windows Commands for Windows 10 would involve no new 
    content.

    Start with the main Windows Commands page for Windows 
    Server 2016. Delete all the links to commands that exist  
    only on Windows Server. Store the page in the Windows 10 
    documentation hierarchy. You now have documentation of 
    Windows Commands for Windows 10. At least on the Web.

    I do not know what generating a PDF version entails. And I do not 
    care. After years of no current documentation of Windows 
    Commands in PDF format. AND no documentation of Windows 
    Commands on the Web, I would settle for current documentation 
    of Windows Commands on the Web.


2) For the MORE command

   a) As Yetoo1 pointed out, there is no explanation of what the /E 
       option does

   b) (Remarks section. Bullet Point 1. Part after 1st sentence.) The 
       word and symbol order lead me to believe that somebody 
       intended for this part to be a table. And did not generate the 
       table correctly. This part certainly would be clearer as a table


The MORE page describes the extended behavior that /E gives you. But not anywhere associated with /E. In the Remarks section at Bullet Point 1.

On my copy of Windows 10, MORE with /E allows the options listed in the Remarks section. And MORE WITHOUT /E allows the options listed in the Remarks section. The "extended" commands work regardless of the /E option.

[...]

PR #6085, opened 2021-11-25
Add /e parameter to better reflect actual more help text

Yetoo1:
Documentation without /e parameter doesn't reflect actual command despite /e not having functional effect. Many user tests indicate there couldn't be functional effect, but no Microsoft employee has definitively clarified that /e has no effect. Even if it truly was that the parameter has no functional effect, it must be included in the documentation so people will know that passing /e is a valid command and will not yield an invalid parameter error. Modifies usage syntax and parameter table and subcommand remark to better accommodate /e.

Addresses #2016.

### DIFF
--- a/WindowsServerDocs/administration/windows-commands/more.md
+++ b/WindowsServerDocs/administration/windows-commands/more.md
@@ -19,9 +19,9 @@ Displays one screen of output at a time.
 ## Syntax
 
 ```
-<command> | more [/c] [/p] [/s] [/t<n>] [+<n>]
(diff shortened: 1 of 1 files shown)


PRMerger13:
@Yetoo1 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Yetoo1:
#sign-off

PRMerger14:
@Yetoo1: I'm sorry - only the author of this article, @JasonGerend, can sign off on your changes. But we do have an exception process - if you are on the Microsoft content or product team for this product area, you can ask the PR review team to review and merge it by sending mail to the techdocprs alias.

PR #6118, opened 2021-12-13
Added paranthesis

jwmoss:
Added parenthesis around the format-trend output. Without it, only +/day is returned, not the value.

image


PRMerger18:
@jwmoss : Thanks for your contribution! The author(s) have been notified to review your proposed change. @cosmosdarwin

jwmoss:
Ha! Better late than never. Thanks for merging!

PR #6161, opened 2022-01-03
RDWebClientManagement now requires -AcceptLicense parameter to install

jimitkg:
RDWebClientManagement now requires -AcceptLicense parameter to install


PRMerger15:
@jimitkg : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6190, opened 2022-01-13
Update input-method-editor-error.md

tyadon:
This issue was fixed in 9C and 10B. However, the requesting customer has asked that we update documentation to reflect that the issue was addressed in Windows 10+ and remains in Windows 8.1 and below.


PRMerger7:
@tyadon : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6256, opened 2022-02-06
Remove 2016 from the title

Orin-Thomas:
2016 can be removed from the title as this applies to all supported versions of Windows Server with Desktop experience. It also reduces the likelihood that a reader encountering the article will assume that the advice is not valid for WS 2019 and WS 2022


PRMerger9:
@Orin-Thomas : Thanks for your contribution! The author(s) have been notified to review your proposed change. @nirb

Orin-Thomas:
I talked with Artem about it and he said the 2016 was in the title as that's the only version that's actually been tested. So cancelling the request

PR #6257, opened 2022-02-06
Remove refs to 2012 unless appropriate

Orin-Thomas:
This article's original text has 2012 flavor but it applies to all versions since then (2012, 2012 R2, 2016, 2019, 2022). Reworded to apply to all supported versions of Windows Server.


PRMerger4:
@Orin-Thomas : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6294, opened 2022-02-22
Reduce the number of times LockDown is mentioned in the documentation

PeterSmithRedmond:
Per recommendation, we are moving away from recommending LockDown; it's not appropriate for many customers.


PRMerger20:
@PeterSmithRedmond : Thanks for your contribution! The author(s) have been notified to review your proposed change. @Teresa-MOTIV

PR #6301, opened 2022-02-24
Update OpenSSH key management

jeremy-morren:
The instructions as provided for Administrators do not work (see https://stackoverflow.com/a/64868357/6614154). Adding '/remove "Authenticated Users"' to the icals.exe command does seem to fix the problem.


PRMerger8:
@jeremy-morren : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jeremy-morren:
What is the status on this PR?

PR #6334, opened 2022-03-09
Update call.md

npocmaka:
you can call a label like CALL ::LABEL  - with double semicolon also.


PRMerger4:

PR #6340, opened 2022-03-10
Update PowerShell to configure WIA for Chrome

brainpr:
Remove close parenthesis to fix never-match to actual user agent string sent by Chrome browser which contains additional characters after OS name before close parenthesis.


PRMerger8:
@brainpr : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6360, opened 2022-03-21
Update extended-security-updates-overview.md

maryamrai:
The not on Azure - end date for ESU is 10 oct 2023, which is not correctly reflected in the table.


PRMerger7:
@maryamrai : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6372, opened 2022-03-28
Add FAQ for bitlockered CSV questions

cpuck:
Need information on encryption times and if there will be a performance degradation or outage while the encryption occurs?  See related FAQ here that could be repurposed: https://docs.microsoft.com/en-us/windows/security/information-protection/bitlocker/bitlocker-deployment-and-administration-faq#how-long-will-initial-encryption-take-when-bitlocker-is-turned-on


PRMerger20:
@cpuck : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6404, opened 2022-04-11
Update kerberos-constrained-delegation-overview.md

Orin-Thomas:
Updating versions as text only referred to 2012/2012 R2 whereas this applies to all supported versions of Windows Server.


PRMerger15:
@Orin-Thomas : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6414, opened 2022-04-14
Fix refsutil located path

tksh164:
Fix the refsutil located path.

image


PRMerger4:
@tksh164 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6418, opened 2022-04-16
In my actual operation on the windows server 2012 R2 server, I found …

chen1749144759:
…that the document describes the parameters of adding users to groups incorrectly

In my actual operation on the windows server 2012 R2 server, I found that the document describes the parameters of adding users to groups incorrectly, so that the developer cannot apply this command correctly. I hope you can correct it after the local verification.
Sorry, my English is not good, the spelling may not be very smooth, thanks for the translation software


PRMerger19:
@chen1749144759 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6427, opened 2022-04-20
add Enterprise for Virtual Desktops key

MikeBazMSFT:
source: https://github.com/SystemRage/py-kms/blob/master/py-kms/KmsDataBase.xml


PRMerger16:
@MikeBazMSFT : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MikeBazMSFT:
self-closing because apparently although those keys have been "leaked" they are not officially shared.

PR #6430, opened 2022-04-26
params need quotes

jqknono:
"auditpol /list /subcategory:" cannot run with spaces in params, quotes are needed. 
Adding quotes, tested OK on Win10.


PRMerger13:
@jqknono : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6441, opened 2022-05-04
Doskey $B/$b is equivalent to |, not to ( nor *

louiswins:
Looks like there was probably some weirdness when converting the docs to Markdown because pipe is a special character. Escaping the pipe _inside_ a code block seems iffy but it worked in the preview.


PRMerger20:
@louiswins : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6450, opened 2022-05-09

daniel-cow:
auditing server 2008 R2 requires a registry change which wasn't mentioned. This change adds which key needs modified and that 2008 R2 doesnt support the Powershell.


PRMerger9:
@daniel-cow : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6574, opened 2022-06-26
Corrected syntax block

A-Mofi:
The syntax of command **CALL** should be explained with two lines because of it is either possible to call another batch file from within a batch file or call command lines in currently processed batch file below a label in a new batch file context. It is not possible to call command lines below a label in a different a batch file as the syntax block with just the single line lets readers think. The usage help output on running `call /?` in a command prompt window explains the syntax also with two separate lines which avoids any misunderstanding on syntax for usage of command **CALL**.


PRMerger7:
@A-Mofi : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6593, opened 2022-07-08
Fix the default value for "use multimon"

treeleung:
The correct default value for "use multimon" should be 0, from the client's perspective.


PRMerger7:
@treeleung : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Heidilohr (approved):

PR #6684, opened 2022-08-25
Update 2-configure-wsus.md

vpaulreed:
UUP content will come via *.delivery.mp.microsoft.com.  Adding those to the list of firewall URLs.


vpaulreed:
Hi @JasonGerend - can you give me an eta for the changes.

robinharwood:
Hi @vpaulreed,

Now we have these new entries, can we remove line 85: - https\://dl.delivery.mp.microsoft.com? As the wildcard would seemingly include entry.

Thanks,

Robin

robinharwood:
Confirmed with @vpaulreed, both entries should be present for the moment.

PR #6810, opened 2022-10-13
Request for co-ownership of this page

kkkashan:
I am PM in Linux Systems Group and own the distro relationship. Hyper-V support pages have old linux distro releases those are EOL already. This page need to be updates as per latest supported distro releases. AS LSG PM will able to update and maintain this page.


kkkashan:
submitting again.

PR #6868, opened 2022-11-17
fix EAP server validation documentation and sync sections

Nothing4You:
This fixes and syncs up the sections for `Connect to these servers` in EAP configuration.

The first two sections didn't make any sense at all without a `*` being mentioned there.

Unfortunately, even with the documentation derived from the third section, the original content is still not plausible.
As I don't know the internals, I don't know if a `*` is the only possible wildcard, as it's technically not a wildcard in regex.

The configuration in the GPO provides the following example for server names: `srv1;srv2;.*\.srv3\.com`

`nps*.example.com` does not make much sense in this context, as this regex would match e.g. `npssssaexample.com` but not `nps1.example.com`.
Dots are single character wildcards in regex they must be escaped when a literal dot is intended to avoid unexpected behavior, asterisks allow zero or more of the preceding character.

To match `nps1.example.com` and `nps2.example.com` one option would be `nps.\.example\.com`, but this wouldn't contain an asterisk.
Using an asterisk as the only possible regex "flag" doesn't make much sense, that doesn't mean it's not implemented that way though.

### DIFF
--- a/WindowsServerDocs/networking/technologies/extensible-authentication-protocol/network-access.md
+++ b/WindowsServerDocs/networking/technologies/extensible-authentication-protocol/network-access.md
@@ -75,7 +75,7 @@ This item specifies that the client verifies that server certificates presented 
 
 ### Connect to these servers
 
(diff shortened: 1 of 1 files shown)


khdownie:
@Nothing4You thanks for this. I'm no longer the owner of this article, nor do I work on the Windows Server team. I worked on this a while back as a vendor on a team led by @JasonGerend - perhaps he can find the right person to approve this PR (and probably should also take my name off the article).

jborsecnik:
@Xelu86, @robinharwood, 

IMPORTANT: When the changes are ready for publication, add a `#sign-off` comment to signal that the PR is ready for the review team to merge. 

#label:"aq-pr-triaged"

Xelu86:
Hey @Nothing4You,

The authors have been notified and your contribution is greatly appreciated. This doc has this update applied and is going through several changes which will be reflected during the next business day.

Thank you and have a most pleasant day.
#please-close

PR #6916, opened 2022-12-13
Add 'Additional Notes' for NTLM Block error

vbrennydossa:
When the certificate replacement is executed by an account that is member of 'Protected Users', the NTLM authentication is blocked (and it's used by the replacement). This is documented in the Protected Users page, but not in the certificate replacement page. Currently, the only occurrence of the scenario is under a forum post.  Updating here because it would have saved me 3 days of troubleshooting in a recent case, and may save others :)


Jak-MS:
#assign: @robinharwood @Xelu86 

Can you review the proposed changes? New content should be submitted via the private repo to benefit from staging, validation, and Acrolinx checks. Let us know if you'd like assistance with moving the commits into the private repo.
Note that the file name needs to be revised to meet the file naming guidelines and the new article should be added to a TOC.

#label:"aq-pr-triaged"

PR #6920, opened 2022-12-14
ADFS 2016 - Correct claim type URIs for multi MFA choice 

piaudonn:
All the claim rules examples are incorrect because of incorrect claim type URIs. They were starting with `https://...` instead of `http://...`. Note that although they look like URL, they are not. They are only unique identifiers for the claim types using a URI format. They do not have nor need an `s`.

### DIFF
--- a/WindowsServerDocs/identity/ad-fs/overview/whats-new-active-directory-federation-services-windows-server.md
+++ b/WindowsServerDocs/identity/ad-fs/overview/whats-new-active-directory-federation-services-windows-server.md
@@ -127,7 +127,7 @@ AD FS already supports triggering additional authentication based on claim rule 
 For example, 2012 R2 onwards admin can already write the following rule to prompt additional authentication if the request comes from extranet. 
 
 ```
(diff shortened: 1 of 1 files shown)


BakkerJan:
Ah, that's why the claim rule did not work! This will prevent a lot of troubleshooting for many customers 🙏🏻

Jak-MS:
#assign: @robinharwood 
#assign: @Xelu86 

Can you review the proposed changes? New content should be submitted via the private repo to benefit from staging, validation, and Acrolinx checks. Let us know if you'd like assistance with moving the commits into the private repo.
Note that the file name needs to be revised to meet the file naming guidelines and the new article should be added to a TOC.

#label:"aq-pr-triaged"

billmath:
Thank you for submitting this issue.  Unfortunately, due to high volumes I was unable to respond to this in a timely manner.  At this point, I am going to close this due to it's age and the possibility that it may have been addressed.  If this is still a problem, please consider submitting a new issue.  Thank you.

#please-close

PR #7039, opened 2023-02-08
Tested CVLK 2019 would also activate Windows 11

justin-cyhuang:
Per tested we have confirmed that CVLM 2019 could be used to activate Windows11.  Customers are confused by this document and we kept getting questions.

Is that possible we update document to reflect the actual behavior?

We are also asking CU to test 2016, if worked I will create another pull request.


Jak-MS:

@robinharwood @Xelu86 

#label:"aq-pr-triaged"

PR #7047, opened 2023-02-10
Correct rndPass usage

pearj:
While testing this on Windows Server 2019, I discovered that `+rndPass` shouldn't be combined with `/pass`, it is supposed to be used on it's own. Otherwise `+rndPass` becomes the actual password, which is unexpected and might not meet password complexity requirements.

Running a `ktpass /?` shows the that options `+rndPass` and `/pass` are separate from each other.

### DIFF
--- a/WindowsServerDocs/administration/windows-commands/ktpass.md
+++ b/WindowsServerDocs/administration/windows-commands/ktpass.md
@@ -23,7 +23,7 @@ ktpass
 [/princ <principalname>]
 [/mapuser <useraccount>]
 [/mapop {add|set}] [{-|+}desonly] [/in <filename>]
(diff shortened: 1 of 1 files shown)


pearj:
In addition to this issue, I have noticed that command line option that starts with a `/` is also allowed to start with a `-` according to `ktpass /?`. Also the capitalisation of some of the commands has been lost too, eg: rndPass, minPass, maxPass, DesOnly, etc.

Court72:
@robinharwood, @Xelu86

Can you review the proposed changes? *IMPORTANT*: When the changes are ready for publication, add a `#sign-off` comment to signal that the PR is ready for the review team to merge. 

#label:"aq-pr-triaged"

PR #7073, opened 2023-02-24
Update to Latest Max Cluster Size

strecz:
Simple:
Change "Formatting requirements for large files" section's AUS recommendation from 64KB to 2048 KB .

Detail:
In https://learn.microsoft.com/en-us/windows-server/storage/file-server/ntfs-overview, under section "Formatting requirements for large files," it says to set "a 64 KB NTFS allocation unit size." Per "Support for large volumes" section, 64 KB was the "earlier maximum" and "on Windows Server 2019 and newer and Windows 10, version 1709 and newer"... "NTFS can support volumes as large as 8 petabytes" by using, per table in that section, a max cluster size of 2048 KB (max size). Therefore, in the "Formatting requirements for large files" section, where it says "there are new recommendations for formatting volumes," it would be best to state the recommendation for the latest OS is to use the PowerShell parameter "-AllocationUnitSize 2048KB," and not the old "-AllocationUnitSize 64KB," unless backward compatibility is needed.

Addendum:
I believe the "Cluster size recommendations for ReFS and NTFS" page I referenced at https://techcommunity.microsoft.com/t5/storage-at-microsoft/cluster-size-recommendations-for-refs-and-ntfs/ba-p/425960 requires similar corrections for consistency. If the above proposed change is agreed to, then I would also like to update there as well.

Sincerely,
Steven T. Reczek

### DIFF
--- a/WindowsServerDocs/storage/file-server/ntfs-overview.md
+++ b/WindowsServerDocs/storage/file-server/ntfs-overview.md
@@ -54,20 +54,22 @@ To allow proper extension of large .vhdx files, there are new recommendations fo
 
 |Parameter|Description|
 |---|---|
-|-AllocationUnitSize 64KB|Sets a 64 KB NTFS allocation unit size.|
+|-AllocationUnitSize 2048KB|Sets a 2048 KB NTFS allocation unit size.|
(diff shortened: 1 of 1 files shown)


Jak-MS:

@robinharwood @Xelu86 

#label:"aq-pr-triaged"

PR #7105, opened 2023-03-13
Update Cluster-Affinity.md

PhilipElder:
The New-Object step was in there twice.
Added the # hash so that folks that copy and paste the two lines don't end up with an error. Tweaked the group names to make it easier for new to Cluster PowerShell to pick up on what was being said.


PhilipElder:
Agree

PhilipElder:
@microsoft-github-policy-service agree

PR #7337, opened 2023-06-17
Update parameters as cd path always implies change

crisman:
If there is a path cd will change to it, text likely a copy/paste from drive: line just above it.


Court72:
@robinharwood, @Xelu86

#label:"aq-pr-triaged"

PR #7392, opened 2023-07-18
Update OpenSSH_KeyManagement.md

Weishaupt:
On a localized version of Windows (e.g. German) the `Administrators` group is called something else (e.g. `Administratoren`). Thus `iacls` fails when the remote powershell is executed on such a localized system. Using the SID of the Administrators group fixes this.

### DIFF
--- a/WindowsServerDocs/administration/OpenSSH/OpenSSH_KeyManagement.md
+++ b/WindowsServerDocs/administration/OpenSSH/OpenSSH_KeyManagement.md
@@ -1,7 +1,7 @@
 ---
 title: Key-based authentication in OpenSSH for Windows
 description: Learn about OpenSSH Server key-based authentication, generation and deployment for Windows using built-in Windows tools or PowerShell.
-ms.date: 07/12/2022
+ms.date: 02/14/2024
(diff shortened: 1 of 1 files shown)


Court72:
@robinharwood, @Xelu86

#label:"aq-pr-triaged"

Weishaupt:
@Xelu86 the SID for the Administrator Group is well-known and always the same. Why would you have to look it up?

See: https://learn.microsoft.com/en-us/windows-server/identity/ad-ds/manage/understand-security-identifiers

Xelu86:
Hey @Weishaupt, thank you for this contribution! Localization issues were not considered when this script was initially made, kudos for the awareness. We have made an addition to this doc.

> _@Xelu86 the SID for the Administrator Group is well-known and always the same. Why would you have to look it up?
> See: https://learn.microsoft.com/en-us/windows-server/identity/ad-ds/manage/understand-security-identifiers_

We must be mindful that most users do not know SIDs by heart. Adding the PS command to pull SIDs is a nice detail so users can readily see the current SIDs available to them in their current environment.

We'll be merging these changes shortly.

PR #7436, opened 2023-08-08
Add well-known SID for "Cloud Account Authentication"

cnotin:
This SID isn't widely documented yet, but it's referenced at least here: https://learn.microsoft.com/en-us/troubleshoot/azure/active-directory/no-local-admin-privileges-azure-ad-joined-device

And I observed it when authenticating with an Entra (Azure AD) user on an Entra (Azure AD) joined machine


Court72:
@robinharwood, @Xelu86

#label:"aq-pr-triaged"

PR #7437, opened 2023-08-08
Update cluster-to-cluster-storage-replication.md

conner-s:
Under Step 1.7 the instructions say to install the File Server and storage replica role and feature, but it's not clear if those are both roles or where in each list they are

I defined that File Server was a Role and Storage Replica is a Feature

I also added the missing Failover Clustering Feature that is shown in the PowerShell method of deploying that was not called out in the Graphical instructions


Jak-MS:

@robinharwood @Xelu86 

#label:"aq-pr-triaged"

PR #7458, opened 2023-08-18
Added pre-reqs for dedicated cluster IPs and witness file share

conner-s:
Added prerequisites for the dedicated cluster IPs and the need for a witness file share.

Witness file share is called out later in Step 2, but in terms of going through and configuring, that's likely something you'd want to have beforehand.


nedpyle (approved):

Court72:
@robinharwood, @Xelu86

#label:"aq-pr-triaged"

PR #7496, opened 2023-09-14
Update set_1.md regarding having spaces before or after '=' sign

krishnak-msft:
Having a space before and/or after the '=' sign leads to unwanted behavior. And we should call it out specifically.


Jak-MS:

@robinharwood @Xelu86 

#label:"aq-pr-triaged"

PR #7554, opened 2023-10-24
close unused 3389 ports

gsmitheidw:
In the script on: [Change the listening port in Remote Desktop](https://learn.microsoft.com/en-us/windows-server/remote/remote-desktop-services/clients/change-listening-port) new firewall ports are opened for RDP for both TCP and UDP. 

However these leave the default 3389 port open in the firewall. This is a non-privileged port number (ie: over 1024) which could potentially be used by users of a system to host a rogue application as a pseudo-RDP service and perhaps even harvest incoming data. 

Because the predefined rules cannot be modified, whilst its necessary to create two new firewall rules, it would be prudent to add these two lines to the code to close the predefined rules:

    Disable-NetFirewallRule -DisplayName "Remote Desktop - User Mode (TCP-In)"
    Disable-NetFirewallRule -DisplayName "Remote Desktop - User Mode (UDP-In)"


ShannonLeavitt:

@robinharwood @Xelu86 

#label:"aq-pr-triaged"

PR #7580, opened 2023-11-06
Update fsutil.md

gtworek:
Precisely, in some cases you don't have to be an admin to use fsutil. Made it sound more related to the actual way how the tool works.


Court72:
@robinharwood, @Xelu86

#label:"aq-pr-triaged"

PR #7601, opened 2023-11-15
Update ad-fs-requirements.md

rodiera:
The correct name of the user right assignment permission is "log on as a service" not "Sign-in as a Service"


Xelu86:
Hey @rodiera, we have addressed this issue in another PR and those changes should take affect by next business day. We're closing this PR. Thank you.

PR #7630, opened 2023-12-04
Update ad-fs-requirements.md

moondawggy:
'Sign-in as a Service' needs to be corrected to reflect 'Log on as a Service'


robinharwood:
Great spot on the wording here @moondawggy, merging now.

PR #7685, opened 2024-01-17
Add backticks for commands, args, filenames

carehart:
I'll understand if you may feel these are not needed (or not all needed). I just found that as I was reading the doc, sometimes it wasn't as clear since some commands were not formatted distinctly.

I also appreciate that you may prefer that changes like the above be done only "in whole" across "all related documents at once"--to ensure consistency. But I feared that if I held off proposing this (to await that), I might never have offered even this. :-) I'd also not want to do more if I found you didn't even care for these formatting changes at all.

Finally, I added one minor typo correction while I was at it. Apologies if you'd really prefer that to be in its own commit.


Court72:
@robinharwood, @Xelu86

#label:"aq-pr-triaged"

carehart:
Folks, this seems to have fallen between the cracks, though seemingly agreed to by all concerned back in January.

PR #7705, opened 2024-02-02
Update configuring-additional-lsa-protection.md

chcart:
If RunAsPPL set to 2 does not work on anything except Win 11 22h2 and higher, then it shouldn't be here, since that OS already has its own section.  Can confirm on Win 10 22h2, that setting this to 2 still locks LSAProt in UEFI


ShannonLeavitt:

@robinharwood @Xelu86 

#label:"aq-pr-triaged"

PR #7712, opened 2024-02-07
Update enable-hotpatch-azure-edition.md

Karl-WE:
switched from Invoke-webrequest to Start-BITSTransfer as this is more flexible and offers much better download performance.


Xelu86:
Hey @Karl-WE, thank you for your contribution. We agree with these changes. Merging shortly.

PR #7718, opened 2024-02-09
Update echo.md

jmortiger:
- Added reference to an edge case where using parentheses in an echo statement in a block denoted by parentheses causes the block to incorrectly assume those parentheses are syntax, causing the statement to fail without escaping the parentheses.

- Updated an example to illustrate this in action.

- Updated an example to show the previously missing empty line in the output.


Jak-MS:

@robinharwood @Xelu86 

#label:"aq-pr-triaged"

Xelu86:
Hey @jmortiger, thank you for your contribution! We agree with this addition & changes. Merging shortly.

PR #7723, opened 2024-02-12
secedit-export: The "/db" parameter is optional

exchange12rocks:
On live systems we can clearly see that the "/db" parameter is optional.


Xelu86:
Hey @exchange12rocks, thank you for your contribution! We have modified the description for the `/db` parameter to make it more clear to users.

Merging shortly.

PR #7728, opened 2024-02-13
Update windows-server-2016-improvements.md

JoshMoraDSE:
Through testing, backed up by code review, I've confirmed that GP processing on it's own is not enough to make the w32time service become aware of UtilizeSslTimeData changes, as this value is only ready at service/thread startup. Current wording is a bit misleading, suggests the GP Client will become aware of the change and that alone is enough for changes to take effect, which is not true.


Jak-MS:

@robinharwood @Xelu86 

#label:"aq-pr-triaged"

Jak-MS:
nudge @robinharwood @Xelu86

PR #7812, opened 2024-04-21
fix - Fixed note rendering for redirect printing

AptiviCEO:
We've fixed incorrect note rendering for the "Redirect printing" section. This makes reading the note easier than before.

Without the changes, this is what happens:

Screenshot_20240421_225207_Chrome


Court72:
@robinharwood, @Xelu86

#label:"aq-pr-triaged"

PR #7813, opened 2024-04-21
fix - Fixed iSCSI TSP broken link

AptiviCEO:
We've fixed the iSCSI Target Storage Providers (VDS/VSS) download link starting with `hhttps` when it's supposed to be `https`.

Currently, the link that we've fixed in this PR redirects to the same page.

Screenshot_20240421_230424_Chrome


Court72:
@robinharwood, @Xelu86

#label:"aq-pr-triaged"

PR #7878, opened 2024-05-30

Avatar1976:
Article incorrectly stated that both SMBv1 and SMBv2 status could be determined by reading LanmanWorkstation service entry.  It will show one of the dependencies for LanmanWorkstation service being mrxsmb20 however that does not tell you if that is enabled or not.


Court72:
@robinharwood, @Xelu86

#label:"aq-pr-triaged"

PR #7922, opened 2024-08-08
UpdateSaltParameter

fmata-ms:
Using the current empty value in salt does not allow the connection to wac after install. Adding  salt:  "\"\""  instead of the currrent salt: "" resolves the issue


Court72:
@robinharwood, @Xelu86

#label:"aq-pr-triaged"

PR #7935, opened 2024-10-03
Update Configure-Server-Certificate-Autoenrollment.md

florinbelea:
updated supported OSs


florinbelea:
we also need a reference to this https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2003/cc778245(v=ws.10) in more details

PR #7947, opened 2024-10-11
Update adfs-sql-latency.md

miketheitguy:
- Format cache setting into a code block so it's easy to copy & paste.
- Update example to use recommended 1 hour interval instead of 30 minutes.
- Format additional values to match value formatting used throughout the document.
- Fix minor errors.
- Add note about SQL Service Broker


miketheitguy:
@microsoft-github-policy-service agree

miketheitguy:
Double check my change for the cache line please! Or what the recommended string looks like from PG. The original "note" string would never work, and permutations of it would not allow adfssrv to start. Once I formatted it like I mentioned, the service started.

Jak-MS:

@robinharwood @Xelu86 

#label:"aq-pr-triaged"

PR #7969, opened 2024-11-03
Update icacls with PowerShell usage example

Ferulaindig:
The original document made no mention of syntax formatting using PowerShell. It also contains explicit examples that when using inside PowerShell will return an error.


ShannonLeavitt:

@robinharwood @Xelu86 

#label:"aq-pr-triaged"

PR #7975, opened 2024-11-05
Adding couple recent new builtin groups

cnotin:
Observed in Windows 11 and Windows 2025 (with AD domain installed) using:

```powershell
get-adobject -filter * -Properties objectsid,description | where {$_.objectsid -match "s-1-5-32"} | select name,objectsid, description| Sort-Object objectsid | fl

get-localgroup | select name,description,sid | sort-object sid | fl
```


cnotin:
There are other docs that mention these that need to be updated too. Could you please do it?
* https://learn.microsoft.com/en-us/windows-server/identity/ad-ds/manage/understand-security-groups
* https://learn.microsoft.com/en-us/windows-server/identity/ad-ds/manage/understand-security-identifiers
* https://learn.microsoft.com/en-us/windows/win32/secauthz/well-known-sids
* https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/81d92bba-d22b-4a8c-908a-554ab29148ab

Jak-MS:

@robinharwood @Xelu86 

#label:"aq-pr-triaged"

PR #7980, opened 2024-11-09
Update upgrade-overview.md

k-takai:
Removed Windows Server 2025 from the "upgrade from" row in the supported "upgrade from and to" table.

It is misleading to state that an in-place upgrade path from Windows Server 2025 to Windows Server 2025 is supported.
Regarding upgrades from Windows Server 2025, it is necessary to indicate whether future versions will be supported, and I think it is not necessary to list Windows Server 2025, which is the latest version at this time, as the upgrade from.


Court72:
@robinharwood, @Xelu86

#label:"aq-pr-triaged"

PR #7996, opened 2024-11-27
Update delegated-managed-service-accounts-set-up-dmsa.md

zjorz:
the original WARNING statement, gives you the impression it is possible to delete the original service account, which appears not to be correct. Deleting the original service account, and then restarting the service still using it, fails to start. Undeleting the original service account and starting the service allows logon to then retrieve the password of the dMSA replacing the original service account


Court72:
@robinharwood, @Xelu86

#label:"aq-pr-triaged"

PR #8001, opened 2024-12-07
Update Appendix-G--Securing-Administrators-Groups-in-Active-Directory.md

Air-Git:
The document says "There should be no day-to-day user accounts in the Administrators group with the exception of the Built-in Administrator account for the domain". But it also says "Additionally, DAs and EAs inherit a number of their rights and permissions by virtue of their default membership in the Administrators group. Default group nesting for privileged groups in Active Directory should not be modified".

This is unclear and possibly contradictory. The term "day-to-day user accounts" is not defined or standard. Does it include the Administrator account and the DA and EA groups? If it does, then the part about not modifying nesting groups is contradictory. If it does not, then the Administrator account cannot be excluded, because it is not included.

It is not clear whether the intention is to remove ALL members, with the possible exception of the Administrators account; or to remove ALL members EXCEPT the default members.

### DIFF
--- a/WindowsServerDocs/identity/ad-ds/plan/security-best-practices/Appendix-G--Securing-Administrators-Groups-in-Active-Directory.md
+++ b/WindowsServerDocs/identity/ad-ds/plan/security-best-practices/Appendix-G--Securing-Administrators-Groups-in-Active-Directory.md
@@ -15,7 +15,7 @@ ms.topic: article
 
 
 ## Appendix G: Securing Administrators Groups in Active Directory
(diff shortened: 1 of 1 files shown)


Air-Git:
Additionally, the text and screenshot refer to removing all members. I have not proposed a change to that text because it is not clear what the recommendation is. My own reading is that "no day-to-day user accounts" means no custom (non-default) accounts and, by extension, any groups of them. But it could also mean "no accounts or groups at all, normally" i.e. except in build or disaster recovery situations. 

It's the same with Appendix F.

ttorble:

@robinharwood @Xelu86  

#label:"aq-pr-triaged" 

PR #8008, opened 2025-01-06
Update azure-hybrid-benefit.md

adturner:
Update language on use rights for Windows Server Datacenter edition - it is not clear with the current verbiage that this is a migration allowance, in conflict with the product terms available here - https://www.microsoft.com/licensing/terms/productoffering/MicrosoftAzure/EAEAS - as written, the statement infers that a datacenter license with software assurance does not need to be tied to a workload that is being migrated, which allows the same license to be used for two explicitly different purposes on-prem and in Azure.


Jak-MS:

@robinharwood @Xelu86 

#label:"aq-pr-triaged"

PR #8023, opened 2025-02-04
Update configure-virtual-local-area-networks-for-hyper-v.md

takayamm:
Our customer pointed out that the current description makes it seem as if the VLAN ID set on the virtual switch is applied to all traffic passing through the physical NIC. In reality, when different VLAN IDs are set on the virtual switch and the virtual NIC on the virtual machine, each VLAN ID is applied accordingly. We have considered a revision to make this point clearer, so please review it.


ttorble:

 

@robinharwood @Xelu86  

#label:"aq-pr-triaged" 

PR #8026, opened 2025-02-06
rundll32: fix documentation and add more detail

nir9:
Problem: Current rundll32 documentation contains incorrect information and lacks information about the syntax.

Example: Currently, the doc claims that rundll32 loads only 32bit DLLs which is untrue, I have tested with a 64bit Windows 10 machine and when rundll32 is ran regularly is loads 64bit DLLs and when it is ran through SysWow64 like so for example ```C:\Windows\SysWow64\rundll32 kernel32,Beep``` it of course uses 32bit DLLs

Here is a screenshot from procexp that shows it uses the 64bit DLL:
image

Solution:
- Remove 32bit only claim
- Remove claim that it requires admin rights (not needed necessarily)
- Add full syntax information and additional remarks

### DIFF
--- a/WindowsServerDocs/administration/windows-commands/rundll32.md
+++ b/WindowsServerDocs/administration/windows-commands/rundll32.md
@@ -11,26 +11,38 @@ ms.date: 10/16/2017
 
 # rundll32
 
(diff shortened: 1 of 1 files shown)


v-dirichards:
@robinharwood, @Xelu86
 
#label:"aq-pr-triaged"

PR #8034, opened 2025-03-02
Update delegated-managed-service-accounts-set-up-dmsa.md

MaxWelzig:
The previous snippet with the property, "Properties", cannot be run in PowerShell. 

It results in the following error: "Set-ADServiceAccount : A parameter cannot be found that matches parameter name 'Properties'."

The new change replaces "Properties" with "Replace". This allows you to run the snippet and update the "msDS-DelegatedMSAState" value from 0 to 3.

Server used: Windows Server 2025 24H2 
OS Build: 26100:1742


Court72:
@robinharwood, @Xelu86

#label:"aq-pr-triaged"

PR #8035, opened 2025-03-04
Add Exchange connectivity warning for NCSI setting.

The00Dustin:
I have isolated an issue regarding Outlook connecting to an on-prem standalone Exchange server to this particular policy setting.  This commit adds a warning to the comments section of that policy setting.  This warning might benefit from additional emphasis or detail.  With regard to additional detail, I do not have O365 or M365 to compare connectivity to a cloud server or hybrid setup vs an on-prem standalone setup  or to validate whether this would also affect Classic Outlook or New Outlook.

The issue is reproducible by enabling this setting without any other device settings in the policy and clears when this setting is disabled.  The issue affected connectivity with Exchange 2019 from Outlook 2016 on Windows 10 Enterpise LTSC 2021 and also affected connectivity to the same server from Outlook 2021 on Windows 11 Enterprise LTSC 2024.

After this setting is applied, it takes a reboot before the issue surfaces even if the machine was offline when the new policy was set and gets the policy during boot.  Autodiscover will still work, but Outlook will work offline with regard to the Exchange server, no connectivity will even be attempted from Outlook 2016 (nothing in netstat, nothing in fiddler).


v-dirichards:
@robinharwood, @Xelu86
 
#label:"aq-pr-triaged"

meaghanlewis (approved):

meaghanlewis:
@The00Dustin thanks for your contribution!

PR #8043, opened 2025-04-23
Fixed typo of "server"

bigstevo:
"Download update files to this serer only when updates are approved" > "Download update files to this server only when updates are approved"


v-regandowner:
Thanks @bigstevo for your contribution!

PR #8048, opened 2025-05-06
cmd: add documentation of the double pipe feature

nir9:
Problem: Currently the cmd documentation only mentions the `&&` feature as a way to chain multiple commands without going into detail that it only proceeds if the preceding command succeeds. The documentation also lacks the complementary `||`.

Solution: Add more details regarding `&&` and add the missing `||`.


v-dirichards:
@robinharwood, @Xelu86
 
#label:"aq-pr-triaged"

PR #8058, opened 2025-05-30
Update 2-configure-wsus.md

vpaulreed:
We have removed the need to add the domain wustat.windows.com to your firewall as we no longer use that url.


Court72:
@robinharwood, @Xelu86

#label:"aq-pr-triaged"

PR #8071, opened 2025-07-03
Removed "Set-SmbServerConfiguration -EnableInsecureGuestLogons" command

omry99:
`EnableInsecureGuestLogons` is not a parameter of `Set-SmbServerConfiguration` (only of `Set-SmbClientConfiguration`), as can  be seen [here ](https://learn.microsoft.com/en-us/powershell/module/smbshare/set-smbserverconfiguration?view=windowsserver2025-ps)


robinharwood:
Thanks for your contribution @omry99.

PR #8101, opened 2025-11-21
Remove "you can rollback by just reinstalling" messaging.

chcart:
Removed 
"The CA database, private key, and certificate aren't removed from the source server by removing the CA role service. Therefore, reinstalling the CA role service on the source server restores the source CA if migration fails and performing a rollback is required, " as the first sentence is misleading and the second sentence is just completely wrong.  Reinstalling the CA role overwrites the CA database and you lose everything in it.


chcart:
Dunno why this is still sitting here.  Made changes in -pr

PR #8132, opened 2026-04-18
Fix incorrect backslash in dir wildcard syntax example

nir9:
Problem: The command currently presented in the example does not work since the backslash is parsed by CMD as a path separator.

Solution: Remove the backslash from the wildcard example.


ttorble:
#assign: @robinharwood, @dknappettmsft  

 

@robinharwood @dknappettmsft  

#label:"aq-pr-triaged" 

